PR #7, opened 2018-05-07
Feature/functional primitives

muratovv:
Add functional primitives of `map` and `end` for nullable types


src/main/kotlin/util/functional/bind.kt
```diff
@@ -0,0 +1,31 @@
+package util.functional
+
+/**
+ * Functional application of lambda to nullable type
+ * @return application of lambda to value or null
+ */
+fun <T, R> T?.map(map: (value: T) -> R): R? {
```
Alexey-N-Chernyshov: lambda can return nullable type
(value: T) -> R?

PR #8, opened 2018-05-09
Bump to JUnit5

Alexey-N-Chernyshov:
Bump to JUnit5 ⬆️
- Now it is possible to mock final classes
- rework all tests with JUnit5
- rename tests


src/test/kotlin/vendor/EthUsageTest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+class EthUsageTest {
+
+    @Test
+    fun ethUsage() {
+        try {
+            val web3 = Web3j.build(HttpService())
+            web3.web3ClientVersion().observable().subscribe { x ->
+                val clientVersion = x.getWeb3ClientVersion()
+                print(clientVersion)
```
x3medima17: What if client is running?
Should it assert with true or false?

Alexey-N-Chernyshov: This test checks the library is loaded. So, it should always pass.

x3medima17: Maybe it is better to add explicit `assert(true)` after printing client version?

PR #10, opened 2018-05-11
Feature/eth chain handler

Alexey-N-Chernyshov:
Add Ethereum chain handler


src/main/kotlin/sideChain/eth/EthChainHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+
+        val res = mutableListOf<NotaryEvent>()
+        val txs = block.block.transactions
+
+        for (txRes in txs) {
+            val tx = txRes.get() as Transaction
+
+            // we are interested in transactions to a particular address
+            if (tx.to == Configs.ethListenAddress)
+                res.add(NotaryEvent.EthChainEvent.OnEthSidechainTransfer(tx.hash, tx.from, tx.value, tx.input))
```
x3medima17: Maybe this can be done in more functional way?

PR #15, opened 2018-05-14
add implementation of Ethereum deposit

Alexey-N-Chernyshov:
implement deposit eth

Signed-off-by: Alexey Chernyshov <alexey.n.chernyshov@gmail.com>


src/main/kotlin/main/main.kt
```diff
@@ -18,9 +18,9 @@ fun main(args: Array<String>) {
     val notary = NotaryInitialization()
 
     IrohaInitializtion.loadIrohaLibrary()
-            .flatMap { notary.init() }
-            .failure {
-                logger.logger.error { it }
-                System.exit(1)
-            }
+        .flatMap { notary.init() }
```
x3medima17: How do you format the code?

src/main/kotlin/sideChain/iroha/consumer/IrohaConverterImpl.kt
```diff
@@ -19,50 +20,57 @@ class IrohaConverterImpl {
...
+                            )
+                    }
                 }
+                txs.add(txBuilder.build())
             }
-            txs.add(txBuilder.build())
+        } catch (e: Exception) {
+            logger.error { e }
         }
         return txs
```
x3medima17: I think it is better to return `optional` here.

src/main/kotlin/main/main.kt
Alexey-N-Chernyshov: option + cmd + L

Alexey-N-Chernyshov: IntelliJ Idea settings are described here (https://kotlinlang.org/docs/reference/coding-conventions.html#applying-the-style-guide)

src/main/kotlin/sideChain/iroha/consumer/IrohaConverterImpl.kt
Alexey-N-Chernyshov: Good point, but it seems better to handle errors with RxJava2. I'll fix it.

src/main/kotlin/main/main.kt
x3medima17: The point is that I format it the same way, and I get different result :D

PR #20, opened 2018-05-20
Iroha deploy script

x3medima17:

This script runs an iroha node.
Just run `./deploy/iroha/deploy.sh`


deploy/iroha/deploy.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+CURDIR="$(cd "$(dirname "$0")"; pwd)"
+
+docker rm -f iroha
+docker rm -f some-postgres
+docker network create iroha-network || true
+
+docker run --name some-postgres \
```
Alexey-N-Chernyshov: I think it is better to rename to d3ledger-postgres to avoid name clashing.

deploy/iroha/deploy.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+CURDIR="$(cd "$(dirname "$0")"; pwd)"
+
+docker rm -f iroha
```
Alexey-N-Chernyshov: I think it should be renamed to d3ledeger-iroha to avoid name clashing.

deploy/iroha/config.docker
```diff
@@ -0,0 +1,11 @@
+{
+  "block_store_path" : "/tmp/block_store/",
+  "torii_port" : 50051,
+  "internal_port" : 10001,
+  "pg_opt" : "host=d3ledger-postgres port=5432 user=postgres password=mysecretpassword",
+  "max_proposal_size" : 10,
+  "proposal_delay" : 5000,
+  "vote_delay" : 5000,
+  "load_delay" : 5000,
+  "mst_enable" : false
```
Alexey-N-Chernyshov: We need mst, but we can enable it later if it is not tested yet well.

PR #21, opened 2018-05-21
Notary on Ethereum deposit

Alexey-N-Chernyshov:
Add logic on Ethereum side chain deposit into Notary.


src/test/notary/NotaryTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        val obsEth = Observable.just<NotaryInputEvent>(custodianIntention)
+        val obsIroha = Observable.empty<NotaryInputEvent>()
+
+        val notary = NotaryImpl(obsEth, obsIroha)
+        val res = notary.irohaOutput()
+
+        res.subscribe(
+            {
+                when (it) {
+                    is IrohaOrderedBatch -> {
```
x3medima17: What if not an `IrohaOrderedBatch` arrives?

Alexey-N-Chernyshov: fixed

PR #22, opened 2018-05-23
Feature/integration test

Alexey-N-Chernyshov:
Add functions for ethereum side chain
- send Ether to the listening address
- query balance of associated Iroha account
- deposit ethereum case


src/test/integration/IntegrationTest.kt
```diff
@@ -0,0 +1,221 @@
...
+import org.web3j.protocol.core.DefaultBlockParameterName
+import org.web3j.protocol.http.HttpService
+import org.web3j.utils.Numeric
+import sideChain.iroha.IrohaInitializtion
+import sideChain.iroha.consumer.IrohaKeyLoader
+import sideChain.iroha.util.toByteArray
+import java.math.BigInteger
+
+/**
+ * Class for Ethereum sidechain infrastructrure deployment and communiation.
```
luckychess: **infrastructure** and **communication**

src/test/integration/IntegrationTest.kt
```diff
@@ -0,0 +1,221 @@
...
+            ).send()
+
+        return contract.contractAddress
+    }
+
+    /**
+     * Deploy notary smart contract
+     * @return notary smart contract address
+     */
+    private fun deployTokenSmartContract(): String {
```
luckychess: I think `deployNotarySmartContract` would be a better name.

src/test/integration/IntegrationTest.kt
```diff
@@ -0,0 +1,221 @@
...
+        println("Notary contract address: $notary")
+        println("User contract address: $user")
+    }
+
+    /**
+     * Test US transfer Ethereum.
+     * Note: Ethereum and Iroha must be deployed to pass the test.
+     * @given Ethereum and Iroha networks running and two ethereum wallets and "fromAddress" with at least 0.001 Ether
+     * (1234000000000000 Wei) and notary running
+     * @when "fromAddress" transfers 100 Wei to "toAddress"
```
luckychess: Looks like it transfers a bit more than 100 Wei :) Also @then is missing.

PR #33, opened 2018-06-05
Feature/on add peer handler

x3medima17:
Added Block Streaming client and server for IrohaChainListener


src/test/notary/IrohaBlockEmitterTest.kt
```diff
@@ -0,0 +1,100 @@
...
+
+
+    val period: Long = 1
+    val unit = TimeUnit.SECONDS
+    val expectedBlocks = 5
+    val timeout = ceil(period * expectedBlocks * 1.2).toLong()
+
+
+    @Test
+    fun emitterTest() {
```
Alexey-N-Chernyshov: please add given-when-then notation

PR #38, opened 2018-06-06
Add backbone of refund

muratovv:
Add pipeline of refund.
There are 3 steps:
* query to Iroha
* refund creation
* and signing
Each step is potentially separated task. Also, in steps implementation, we will separate those steps by own class with dependencies and interfaces.


src/main/kotlin/endpoint/eth/EthNotaryResponse.kt
```diff
@@ -13,7 +13,7 @@ sealed class EthNotaryResponse {
      */
     data class Successful(
         val ethSignature: EthSignature,
-        val ethPublicKey: EthPublicKey
+        val ethRefund: EthRefund
```
Alexey-N-Chernyshov: please, add documentation for data fields

src/main/kotlin/endpoint/eth/EthRefund.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Type of coin in Ethereum. Possibly it will be an address
+ */
+typealias CoinType = String
+
+/**
+ * Type of amount in Ethereum
+ */
+typealias AmountType = Double
```
Alexey-N-Chernyshov: I think amount in ether is BigInteger. It seems to me that it is better to move all Ethereum data types into separate file.

src/main/kotlin/endpoint/eth/EthRefundStrategyImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * The method passes the query with transaction to Iroha and return the answer
+     * @param request - user's request with transaction hash
+     * @return Transaction which is appeared in Iroha or error
+     */
+    private fun performQuery(request: EthRefundRequest): Result<iroha.protocol.BlockOuterClass.Transaction, NotaryException> {
+        val query = ModelQueryBuilder()
+        val hashes = HashVector()
+        HashVector().add(Hash(request.irohaTx))
```
Alexey-N-Chernyshov: mb hashes.add()?

muratovv: What do you mean?

src/main/kotlin/endpoint/eth/EthRefund.kt
muratovv: Yeah, we have to investigate the type on eth side and put it here.

src/main/kotlin/notary/NotaryInitialization.kt
```diff
@@ -142,16 +143,13 @@ class NotaryInitialization {
...
             ServerInitializationBundle(CONFIG[ConfigKeys.refundPort], CONFIG[ConfigKeys.ethEndpoint]),
             mock {
-                val request = any<EthRefundContract>()
+                val request = any<EthRefundRequest>()
                 on {
                     performRefund(request)
                 } doReturn EthNotaryResponse.Successful(
                     "signature",
-                    "pub_key"
+                    EthRefund("address", "coin", 66.6)
```
Alexey-N-Chernyshov: Currently amount is BigInteger, so `BigInteger.valueOf(666)`

muratovv: The issue has been fixed already.

PR #44, opened 2018-06-13
Feature/db jooq

Alexey-N-Chernyshov:
Add whitelists for ERC20 tokens and wallets with jOOQ DB layer.


deploy/db/sql/populate.sql
```diff
@@ -0,0 +1,10 @@
+-- Populate wallets table
+INSERT INTO notary.wallets (wallet, irohausername)
+VALUES ('0x00aa39d30f0d20ff03a22ccfc30b7efbfca597c2', 'user1@notary');
```
x3medima17: Maybe it is better to store all these addresses in a separate config file?

src/test/integration/EthTokensProviderTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Test Postgres ethereum ERC20 tokens provider
+ */
+class EthTokensProviderTest {
+
+    /**
+     * List all tokens in database
+     */
+    //    @Test
```
x3medima17: Looks like it is not a test

src/test/integration/EthTokensProviderTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        )
+
+        DSL.using(connection).use { ctx ->
+            val tokens = Tokens.TOKENS
+
+            // read
+            ctx.select(tokens.WALLET, tokens.TOKEN)
+                .from(tokens)
+                .forEach {
+                    println(it)
```
x3medima17: No need in println in tests

src/test/integration/EthWalletsProviderTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Test Postgres ethereum wallets provider
+ */
+class EthWalletsProviderTest {
+
+    /**
+     * List all wallets in database
+     */
+    //    @Test
```
x3medima17: sab

src/test/integration/EthWalletsProviderTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        )
+
+        DSL.using(connection).use { ctx ->
+            val wallets = Wallets.WALLETS
+
+            // read
+            ctx.select(wallets.WALLET, wallets.IROHAUSERNAME)
+                .from(wallets)
+                .forEach {
+                    println(it)
```
x3medima17: sab

deploy/db/sql/populate.sql
Alexey-N-Chernyshov: It is testing data, not configuration. These data is used for manual and integration testing.

src/test/integration/EthTokensProviderTest.kt
Alexey-N-Chernyshov: It's a way to disable test in CI, but run it manually.

PR #55, opened 2018-06-29
add relay provider

Alexey-N-Chernyshov:
add relay provider that reads free relays from master notary account and return first one

### DIFF
--- a/src/main/kotlin/registration/EthFreeWalletsProvider.kt
+++ b/src/main/kotlin/registration/EthFreeWalletsProvider.kt
@@ -17,20 +17,31 @@ import java.math.BigInteger
 
 /**
  * Provides with free ethereum relay wallet
+ * @param keypair - iroha keypair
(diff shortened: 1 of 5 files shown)


src/main/kotlin/registration/EthFreeWalletsProvider.kt
```diff
@@ -1,15 +1,74 @@
...
-class EthFreeWalletsProvider {
+class EthFreeWalletsProvider(val keypair: Keypair) {
 
-    // TODO - a.chernyshov - D3-106 -  add effective implementation
-    // Possible solutions:
-    // 1) set listener and update Postgres DB
-    // 2) query Iroha each time
     fun getWallet(): String {
-        return "eth_wallet"
+        val creator = CONFIG[ConfigKeys.irohaCreator]
```
luckychess: May be add test for ```getWallet()```?

PR #56, opened 2018-07-02
add correct genesis block and change configs

Alexey-N-Chernyshov:
Add Iroha genesis block with roles.
Fix configs.


deploy/iroha/genesis.block
```diff
@@ -13,74 +13,71 @@
...
+                        "roleName":"client",
+                        "permissions":[
+                            "can_transfer",
+                            "can_receive"
+                        ]
+                    }
+                  },
+                  {
+                    "createRole":{
+                        "roleName":"none",
```
x3medima17: 👍

src/test/integration/IntegrationTest.kt
```diff
@@ -46,18 +46,18 @@ class IntegrationTest {
...
 
     /** Iroha port */
     val irohaPort = CONFIG[ConfigKeys.irohaPort]
 
     /** Iroha transaction creator */
-    val creator = CONFIG[ConfigKeys.irohaCreator]
+    val creator = CONFIG[ConfigKeys.testIrohaAccount]
+
+    /** Iroha keypair */
+    val keypair: Keypair =
```
x3medima17: Do we really need to specify the type?

x3medima17:
Please do something with config, because I can't merge block streaming 
And I can't merge your branch into mine because it has new genesis block

PR #57, opened 2018-07-02
rewrite deposit ETH test

Alexey-N-Chernyshov:
fix deposit ETH test


src/test/integration/IntegrationTest.kt
```diff
@@ -278,22 +283,30 @@ class IntegrationTest {
      */
     @Disabled
     @Test
-    fun runMain() {
+    fun depositOfETH() {
```
luckychess: Why it's still disabled?

PR #77, opened 2018-07-17
Feature/refactor konfig

Alexey-N-Chernyshov:
Replace konfig with cfg4k.
Now config is no more static object, it is a composed object that is created in every service `main` and is passed to every submodule.


src/main/kotlin/config/Configs.kt
```diff
@@ -0,0 +1,46 @@
...
+    val url: String
+    val username: String
+    val password: String
+}
+
+/**
+ * Load configs from Java properties
+ */
+fun <T : Any> loadConfigs(prefix: String, type: Class<T>): T {
+    val loader = PropertyConfigLoader(ClasspathConfigSource("/defaults.properties"))
```
luckychess: Will it work with CI configuration?

src/main/kotlin/sidechain/eth/util/Crypto.kt
```diff
@@ -12,13 +11,13 @@ import java.math.BigInteger
...
+fun signUserData(ethereumConfig: EthereumConfig, toSign: String): String {
     // TODO luckychess 26.06.2018 D3-100 find a way to produce correct signatures locally
-    val deployHelper = DeployHelper()
-    val parity = Parity.build(HttpService(CONFIG[ConfigKeys.testEthConnectionUrl]))
+    val deployHelper = DeployHelper(ethereumConfig)
+    val parity = Parity.build(HttpService(ethereumConfig.url))
     parity.personalUnlockAccount(
         deployHelper.credentials.address,
-        CONFIG[ConfigKeys.relayRegistartionEthCredentialPassword]
+        ethereumConfig.credentialsPassword
```
luckychess: May be it would be better to load address from config as well and get rid of ```DeployHelper``` in this function?

src/test/kotlin/notary/NotaryTest.kt
```diff
@@ -105,8 +114,8 @@ class NotaryTest {
     @Test
     fun depositEthereumTest() {
         val expectedAmount = BigInteger.valueOf(100)
-        val expectedAssetId = "ether"
-        val expectedCreatorId = CONFIG[ConfigKeys.notaryIrohaAccount]
+        val expectedAssetId = "eth"
```
luckychess: Why "eth"? It looks like we use "ether" everywhere.

src/main/kotlin/sidechain/eth/util/Crypto.kt
Alexey-N-Chernyshov: I think it is internal representation of Eth, and it is better to use provided interface.

PR #93, opened 2018-07-25
Fix/transfer 0 eth

Alexey-N-Chernyshov:
Check if transaction from ethereum contains 0 amount.


src/test/integration/IntegrationTest.kt
```diff
@@ -308,6 +345,47 @@ class IntegrationTest {
         assertEquals(amount, queryIroha(assetId, clientIrohaAccount))
     }
 
+    /**
+     * Test US-002 Deposit of ETH
```
luckychess: ETH -> ERC-20 tokens.

PR #95, opened 2018-07-26
add list of supported eth tokens

Alexey-N-Chernyshov:
Deploy ERC20 to ropsten.
Add them to list.

### DIFF
--- a/deploy/db/sql/init/populate.sql
+++ b/deploy/db/sql/init/populate.sql
@@ -6,5 +6,14 @@ INSERT INTO notary.wallets (wallet, irohausername)
 VALUES ('0x00Bd138aBD70e2F00903268F3Db08f2D25677C9e', 'user2@notary');
 
 -- Populate tokens table
-INSERT INTO notary.tokens (wallet, token)
-VALUES ('0x00Bd138aBD70e2F00903268F3Db08f2D25677C9e', 'tkn');
+INSERT INTO notary.tokens (wallet, token) VALUES
+('0x9d65d6209bcd37f1f546315171b000663117d42f', 'omg'),
+('0x92fd99502914010942db94181061cdda1ce3b370', 'bat'),
+('0x084f2750c7eb6bb6e40cecfed5f54b198ab1b8c7', 'rep'),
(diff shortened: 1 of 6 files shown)


deploy/db/sql/init/populate.sql
```diff
@@ -6,5 +6,14 @@ INSERT INTO notary.wallets (wallet, irohausername)
 VALUES ('0x00Bd138aBD70e2F00903268F3Db08f2D25677C9e', 'user2@notary');
 
 -- Populate tokens table
-INSERT INTO notary.tokens (wallet, token)
-VALUES ('0x00Bd138aBD70e2F00903268F3Db08f2D25677C9e', 'tkn');
+INSERT INTO notary.tokens (wallet, token) VALUES
```
luckychess: What about lines above, do we need them?

src/main/kotlin/sidechain/eth/util/DeployHelper.kt
```diff
@@ -65,16 +65,16 @@ class DeployHelper(ethereumConfig: EthereumConfig, ethereumPasswords: EthereumPa
...
      */
-    fun deployBasicCoinSmartContract(): BasicCoin {
+    fun deployERC20TokenSmartContract(): BasicCoin {
         return contract.BasicCoin.deploy(
             web3,
             credentials,
             gasPrice,
             gasLimit,
-            BigInteger.valueOf(1000),
+            BigInteger.valueOf(Long.MAX_VALUE),
```
luckychess: Do we really need so much tokens?

deploy/db/sql/init/populate.sql
Alexey-N-Chernyshov: Good point, I'll add task to remove PostgresWalletsProvider

src/main/kotlin/sidechain/eth/util/DeployHelper.kt
Alexey-N-Chernyshov: Why not? For example, I want to deploy long term tokens in Ropsten for demo. 1000 is not enough definitely.

PR #98, opened 2018-07-30
move withdrawal integration tests

Alexey-N-Chernyshov:
move withdrawal test to separate file

### DIFF
--- a/src/integration-test/kotlin/integration/WithdrawalIntegrationTest.kt
+++ b/src/integration-test/kotlin/integration/WithdrawalIntegrationTest.kt
@@ -1,0 +1,205 @@
+package integration
+
+import com.github.kittinunf.result.failure
+import com.github.kittinunf.result.flatMap
+import config.EthereumPasswords
+import config.TestConfig
+import config.loadConfigs
+import contract.BasicCoin
+import jp.co.soramitsu.iroha.Keypair
+import jp.co.soramitsu.iroha.ModelCrypto
+import jp.co.soramitsu.iroha.ModelTransactionBuilder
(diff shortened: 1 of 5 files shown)


src/integration-test/kotlin/integration/WithdrawalIntegrationTest.kt
```diff
@@ -0,0 +1,205 @@
...
+     * @param creator - iroha transaction creator
+     * @param kp - keypair
+     * @param srcAccountId - source account id
+     * @param destAccountId - destination account id
+     * @param assetId - asset id
+     * @param description - transaction description
+     * @param amount - amount
+     * @return hex representation of transaction hash
+     */
+    fun transferAssetIroha(
```
luckychess: May be we should just call ```transferAssetIroha``` from ModelUtil?

src/main/kotlin/util/StringExt.kt
```diff
@@ -15,3 +17,13 @@ fun String.hexToAscii(): String {
 
     return output.toString()
 }
+
+/** Returns tandom string of [len] characters */
```
luckychess: Typo in random.

src/integration-test/kotlin/integration/WithdrawalIntegrationTest.kt
Alexey-N-Chernyshov: we can't, there is a custom keypair

Alexey-N-Chernyshov:
We use "squash and merge" strategy, so it will be one commit in master branch.

PR #105, opened 2018-08-04
Ganache-cli

luckychess:
### What is this
This pull request adds ganache-cli to local development environment.

### Why do we need this
All tests for contracts can be now executed in ~30 seconds instead of ~3 hours.

### How to use
Just run `docker-compose -f docker-compose-dev.yml up` and then run contract tests.

### Drawbacks
Some weird code was added due to config problems.

P.S.: only tests in `ContractsTest.kt` file are affected


src/main/kotlin/sidechain/eth/util/DeployHelper.kt
```diff
@@ -17,16 +17,32 @@ import java.math.BigInteger
 /**
  * Helper class for contracts deploying
  */
-class DeployHelper(ethereumConfig: EthereumConfig, ethereumPasswords: EthereumPasswords) {
+class DeployHelper(ethereumConfig: EthereumConfig, ethereumPasswords: EthereumPasswords, etherTest: Boolean = false) {
```
Alexey-N-Chernyshov: Please, add comments for parameters.

PR #110, opened 2018-08-09
New integration test helper class that makes tests truly independent

dolgopolovwork:
New helper class for making independent integration tests. Also roles were ranamed, because it can not contain uppercase letters.


src/main/kotlin/registration/EthFreeWalletsProvider.kt
```diff
@@ -16,29 +15,33 @@ import sidechain.iroha.util.getRelays
...
      */
     fun getWallet(): Result<String, Exception> {
+        return getWallets().map { freeWallets -> freeWallets.first() }
+    }
+
+    /**
+     * Get all free ethereum relay wallets
+     * @return free ethereum relay wallets
+     */
+    fun getWallets(): Result<Set<String>, Exception> {
```
Alexey-N-Chernyshov: may be rename to getFreeRelays()?

src/main/kotlin/notary/EthWalletsProviderIrohaImpl.kt
```diff
@@ -13,15 +13,15 @@ import sidechain.iroha.util.getRelays
...
-    val irohaConfig: IrohaConfig,
-    val keypair: Keypair,
-    val irohaNetwork: IrohaNetwork,
-    val account: String,
-    val setter: String
+    private val irohaConfig: IrohaConfig,
+    private val keypair: Keypair,
+    private val irohaNetwork: IrohaNetwork,
+    private val notaryIrohaAccount: String,
+    private val registrationAccount: String
```
Alexey-N-Chernyshov: In general case, account details can be set not by `registrationAccount`. I think we need more meaningful name, like 'accountDetailSetter' and 'accountDetailHolder'.

src/main/kotlin/registration/EthFreeWalletsProvider.kt
dolgopolovwork: what about renaming the whole class as well?

src/main/kotlin/notary/EthWalletsProviderIrohaImpl.kt
dolgopolovwork: yes that is true. but this is wallets provider class and the only account that can add new wallet to Iroha is registrationAccount.

src/main/kotlin/registration/EthFreeWalletsProvider.kt
Alexey-N-Chernyshov: EthFreeRelayProvider

PR #118, opened 2018-08-17
Whitelisting

Alexey-N-Chernyshov:
Introduce whitelisting to D3 Notary. Whitelist is set to client's details by whitelist_setter account. Whitelist is used to restrict Ethtereum withdrawal addresses.


src/integration-test/kotlin/integration/WithdrawalIntegrationTest.kt
```diff
@@ -202,4 +211,81 @@ class WithdrawalIntegrationTest {
...
+        val clientKeypair = integrationHelper.irohaKeyPair
+
+        val amount = "125"
+        val assetId = "ether#ethereum"
+
+        // add assets to user
+        ModelUtil.addAssetIroha(irohaConsumer, creator, assetId, amount)
+        ModelUtil.transferAssetIroha(irohaConsumer, creator, creator, clientId, assetId, "", amount)
+
+        // make tra
```
x3medima17: I think you missed something here

src/integration-test/kotlin/integration/helper/IntegrationHelperUtil.kt
```diff
@@ -200,10 +203,45 @@ class IntegrationHelperUtil {
...
+        logger.info("client account $name@notary was created")
         return "$name@notary"
     }
 
+    /**
+     * Add Ethrereum addresses to client whitelist, so that she can withdraw only for that addresses
+     * @param clientAccount - client account id in Iroha network
+     * @param addresses - ethereum addresses where client can withdraw her assets
+     */
+    fun setWhitelist(clientAccount: String, addresses: Array<String>) {
```
x3medima17: I would use List instead of Array

src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
```diff
@@ -133,9 +125,30 @@ class EthRefundStrategyImpl(
...
+        return getAccountDetails(
+            irohaConfig,
+            keypair,
+            irohaNetwork,
+            srcAccountId,
+            whitelistSetter
+        ).map { details ->
+            val whitelist = details["eth_whitelist"]
+
+            if (whitelist != null)
```
x3medima17: Koltin should have a better way to do that

src/integration-test/kotlin/integration/WithdrawalIntegrationTest.kt
```diff
@@ -202,4 +211,81 @@ class WithdrawalIntegrationTest {
...
+        async {
+            notary.main(emptyArray())
+        }
+        Thread.sleep(5_000)
+
+        // create client
+        val clientId = integrationHelper.registerClient()
+        val clientKeypair = integrationHelper.irohaKeyPair
+
+        IntegrationHelperUtil().setWhitelist(clientId, arrayOf(toAddress, "qwerqt"))
```
dolgopolovwork: I don't think we need this new instance of IntegrationHelperUtil just to call setWhitelist. You may use integrationHelper  variable instead.

src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
```diff
@@ -133,9 +125,30 @@ class EthRefundStrategyImpl(
         }
     }
 
+    /**
+     * Check if [srcAccountId] has Ethereum withdrawal [address] in whitelist
+     */
+    private fun checkWithdrawalAddress(srcAccountId: String, address: String): Result<Boolean, Exception> {
+        val whitelistSetter = "whitelist_setter@notary"
```
dolgopolovwork: seems like it's hardcoded.

src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
Alexey-N-Chernyshov: Well, explicit null check removes nullable reference and is not so bad in Kotlin. .?let{} operator doesn't allows else branch. If you have better approach, please share it.

x3medima17:
```
whitelist?.let{
    whitelist.split(",").contains(address)
} ?: run {
    false
}
```

Alexey-N-Chernyshov: imho != null is more meaningful

PR #121, opened 2018-08-22
WithdrawalIntegrationTest fix

Alexey-N-Chernyshov:
Fix and rework WithdrawalIntegrationTest.
Now it is working with ganache, pass time reduced.
Next step is to include it to CI pipeline.


src/integration-test/kotlin/integration/helper/IntegrationHelperUtil.kt
```diff
@@ -243,6 +278,37 @@ class IntegrationHelperUtil {
...
+     * @param creator - iroha transaction creator
+     * @param kp - keypair
+     * @param srcAccountId - source account id
+     * @param destAccountId - destination account id
+     * @param assetId - asset id
+     * @param description - transaction description
+     * @param amount - amount
+     * @return hex representation of transaction hash
+     */
+    fun transferAssetIrohaFromClient(
```
luckychess: Could we somehow reuse ```transferAssetIroha``` from ModelUtil? It's quite confusing to have 2 very similar methods in different places.

PR #123, opened 2018-08-23
Vaccuum integration test was fixed

dolgopolovwork:


src/integration-test/resources/ethereum_password.properties
```diff
@@ -1,4 +1,4 @@
 test.credentialsPassword=user
 withdrawal.credentialsPassword=user
 relay-vacuum.credentialsPassword=user
-
+ganache.credentialsPassword=user
```
Alexey-N-Chernyshov: Why do you need it? `test.credentialsPassword` should be used for tests.

Alexey-N-Chernyshov: I think after conflict resolving there is no need to it anymore.

PR #132, opened 2018-08-29
Change signMessage to local sign with web3j

grimadas:
- Add specific format needed for ethereum contracts


src/main/kotlin/sidechain/eth/util/Crypto.kt
```diff
@@ -4,26 +4,31 @@ import config.EthereumConfig
...
  * @return signed data
  */
 fun signUserData(ethereumConfig: EthereumConfig, ethereumPasswords: EthereumPasswords, toSign: String): String {
-    // TODO luckychess 26.06.2018 D3-100 find a way to produce correct signatures locally
-    val deployHelper = DeployHelper(ethereumConfig, ethereumPasswords)
 
-    val admin = Admin.build(HttpService(ethereumConfig.url))
-    admin.personalUnlockAccount(deployHelper.credentials.address, ethereumPasswords.credentialsPassword).send()
-    // TODO: signalize about error if unlock fails
+    val deployHelper = DeployHelper(ethereumConfig, ethereumPasswords)
```
dolgopolovwork: can we move deployHeper to the top level?

grimadas: Pass with params?

grimadas: Actually, we can even remove it now. Good point

src/main/kotlin/sidechain/eth/util/Crypto.kt
```diff
@@ -1,29 +1,31 @@
...
+    val dat = Numeric.hexStringToByteArray(toSign)
+    // Add ethereum signature format
+    val to_sign = ("\u0019Ethereum Signed Message:\n" + (dat.size)).toByteArray()
+    val data_sign = to_sign.plus(dat)
+    val signature = Sign.signMessage(data_sign, ecKeyPair)
+    // Combine in the signature
+    var res = Numeric.toHexString(signature.r)
+    res = res.plus(Numeric.toHexString(signature.s).substring(2))
+    //  The v is always either 27, or 28 - need to convert to 00 or 01
+    res = res.plus("0" + Integer.toString(signature.v.toInt() - 27))
```
dolgopolovwork: what is 27? seems like magic constant. can we create named constant for this?

grimadas: It is Ethereum magic constant, we gonna is only here, so maybe there is no need for naming

PR #134, opened 2018-08-30
Request proofs from all known notaries

luckychess:
Also bug in EthRefundStrategyImpl was fixed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,7 +160,6 @@ test {
 task integrationTest(type: Test) {
     group = LifecycleBasePlugin.VERIFICATION_GROUP
     description = 'Runs the integration tests.'
-    systemProperty 'java.library.path', Paths.get(projectDir.toString(), "iroha_bindings", getCurrentOs()).toString()
 
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
     classpath = sourceSets.integrationTest.runtimeClasspath
--- a/build.gradle
+++ b/build.gradle
@@ -160,6 +160,7 @@ test {
 task integrationTest(type: Test) {
     group = LifecycleBasePlugin.VERIFICATION_GROUP
     description = 'Runs the integration tests.'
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -160,6 +160,7 @@ test {
 task integrationTest(type: Test) {
     group = LifecycleBasePlugin.VERIFICATION_GROUP
     description = 'Runs the integration tests.'
+    systemProperty 'java.library.path', Paths.get(projectDir.toString(), "iroha_bindings", getCurrentOs()).toString()
```
Alexey-N-Chernyshov: Conflict is possible there.

src/main/kotlin/withdrawalservice/WithdrawalServiceImpl.kt
```diff
@@ -87,39 +90,51 @@ class WithdrawalServiceImpl(
...
+
+                    val moshi = Moshi
+                        .Builder()
+                        .add(EthNotaryResponseMoshiAdapter())
+                        .add(BigInteger::class.java, BigIntegerMoshiAdapter())
+                        .build()!!
+                    val ethNotaryAdapter = moshi.adapter(EthNotaryResponse::class.java)!!
+                    val response = ethNotaryAdapter.fromJson(res.jsonObject.toString())
+
+                    if (response is EthNotaryResponse.Error) {
```
Alexey-N-Chernyshov: mb rework with `when`?
```
                   when (response) {
                        is EthNotaryResponse.Error -> {
                            logger.warn { "EthNotaryResponse.Error: ${response.reason}" }
                            return@forEach
                        }
                        
                        is EthNotaryResponse.Successful -> {
                            val signature = response.ethSignature
                            val vrs = extractVRS(signature)
                            vv.add(vrs.v)
                            rr.add(vrs.r)
                            ss.add(vrs.s)
                        }
                    }
```

PR #138, opened 2018-09-03
Feature/prod config

Alexey-N-Chernyshov:
Configs for mainnet with gradle task.
Usage:
`gradle ... -Pprofile=mainnet`


src/main/resources/eth/notary_mainnet.properties
```diff
@@ -0,0 +1,31 @@
+# ========= Notary =========
+eth-notary.registrationServiceIrohaAccount=registration_service_red@notary
+eth-notary.tokenStorageAccount=token_storage_service@notary
+eth-notary.whitelistSetter=whitelist_setter@notary
+# --------- Iroha ---------
+# Iroha peer hostname
+eth-notary.iroha.hostname=http://51.15.84.132:8545
```
luckychess: I believe better to put `localhost` here.

PR #139, opened 2018-09-03
Feature/withdrawal notary list

luckychess:
Finally `getPeerList` is implemented. Still no any sync logic though.


src/main/kotlin/withdrawalservice/WithdrawalServiceConfig.kt
```diff
@@ -9,9 +9,15 @@ interface WithdrawalServiceConfig {
     /** Notary account in Iroha */
     val notaryIrohaAccount: String
 
+    /** Notary storage account in Iroha */
+    val notaryStorageAccount: String
```
Alexey-N-Chernyshov: mb `notaryListStorageAccount`?

PR #149, opened 2018-09-10
Feature/iroha block streaming interface

x3medima17:
Many times we have to wait for exactly one block from iroha, usually we need it in tests.
Using block streaming service and RxJava is an overkill, this PR introduces a suspended function that waits for exactly one block and returns it.

### DIFF
--- a/src/integration-test/kotlin/integration/iroha/IrohaBlockStreamingTest.kt
+++ b/src/integration-test/kotlin/integration/iroha/IrohaBlockStreamingTest.kt
@@ -1,4 +1,4 @@
-package sidechain.iroha
+package integration.iroha
 
 import com.github.kittinunf.result.map
 import config.TestConfig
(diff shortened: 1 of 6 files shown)


src/integration-test/kotlin/integration/iroha/IrohaBlockStreamingTest.kt
```diff
@@ -72,4 +80,42 @@ class IrohaBlockStreamingTest {
...
+
+        val utx = getModelTransactionBuilder()
+            .creatorAccountId(creator)
+            .createdTime(getCurrentTime())
+            .setAccountDetail(creator, "test", "test")
+            .build()
+
+        IrohaConsumerImpl(testConfig.iroha).sendAndCheck(utx)
+
+        runBlocking {
```
Alexey-N-Chernyshov: Why do we need run it blocking?

PR #188, opened 2018-09-26
Fix/notary test

Alexey-N-Chernyshov:
Fix NotaryTest observable
Move PeerListProvider initialisation away from NotaryImpl for better DI.


src/main/kotlin/notary/btc/BtcNotaryInitialization.kt
```diff
@@ -12,14 +12,19 @@ import org.bitcoinj.core.PeerGroup
...
 import sidechain.SideChainEvent
+import sidechain.iroha.util.ModelUtil
 import java.io.File
 
 class BtcNotaryInitialization(
     private val btcNotaryConfig: BtcNotaryConfig,
     private val btcRegisteredAddressesProvider: BtcRegisteredAddressesProvider
 ) {
+
+    val keypair = ModelUtil.loadKeypair(btcNotaryConfig.iroha.pubkeyPath, btcNotaryConfig.iroha.privkeyPath).get()
```
dolgopolovwork: Why get()? Why not map() or at least flatMap()

Alexey-N-Chernyshov: Get extracts value from result in success case and throws an exception if here is failure. It is enough for the test.

PR #203, opened 2018-10-04
Feature/test config

Alexey-N-Chernyshov:
Delete from tests config everything that is done with accountHelper now.
Fix EthRelayProviderTest.
Move IntegrationHelper and AccountHelper to main directory in order to use them from another module.
Remove strange unused method setAccountDetail from ModelUtil.


notary-integration-test/src/main/kotlin/integration/helper/ConfigHelper.kt
```diff
@@ -67,44 +57,37 @@ class ConfigHelper(private val accountHelper: AccountHelper,
         }
     }
 
-    //Creates config for BTC multisig addresses generation
+    /** Creates config for BTC multisig addresses generation */
     fun createBtcPreGenConfig(): BtcPreGenConfig {
+        val btcPkPreGenConfig =
```
dolgopolovwork: why do we need this? now we have to load config from file system every time this function is called.

PR #204, opened 2018-10-04
Split ethereum accounts between services

luckychess:
Each service now has its own ethereum account for local, ropsten and mainnet environments.

Issues with same tx hashes and nonces should appear more rarely but still present.


notary-integration-test/src/integration-test/resources/eth/relay_registration.properties
```diff
@@ -0,0 +1,28 @@
+# ========= Relay registration =========
```
Alexey-N-Chernyshov: We moved this file to 'configs/eth/'

notary-integration-test/src/integration-test/resources/eth/vacuum.properties
```diff
@@ -0,0 +1,27 @@
+# ========= Relay Vacuum =========
```
Alexey-N-Chernyshov: sap

PR #220, opened 2018-10-08
Implement closeable in Eth and Iroha chain listeners

x3medima17:
ChainListener should be closeable because it acquires some resources.
This PR implements this interface for Ethereum and Iroha

### DIFF
--- a/src/main/kotlin/sidechain/ChainListener.kt
+++ b/src/main/kotlin/sidechain/ChainListener.kt
@@ -1,11 +1,12 @@
 package sidechain
 
 import com.github.kittinunf.result.Result
+import java.io.Closeable
 
 /**
  * Class listens for new [Block] in side block chain
  */
-interface ChainListener<Block> {
+interface ChainListener<Block> : Closeable {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/sidechain/ChainListener.kt
```diff
@@ -1,11 +1,12 @@
 package sidechain
 
 import com.github.kittinunf.result.Result
+import java.io.Closeable
 
 /**
  * Class listens for new [Block] in side block chain
  */
-interface ChainListener<Block> {
+interface ChainListener<Block> : Closeable {
```
dolgopolovwork: seems like we have to close listeners in main code too. not tests only.

luckychess: And for other tests too.

x3medima17: There is a separate task for that. I think next sprint might be dedicated to all these kind of stuff like graceful shutdown

PR #222, opened 2018-10-09
Bitcoin CI

dolgopolovwork:
Bitcoin node is included in docker. Now we are able to run bitcoin integration tests in Jenkins.


src/main/kotlin/config/Configs.kt
```diff
@@ -55,6 +55,8 @@ interface BitcoinConfig {
     val blockStoragePath: String
     //Depth of transactions in BTC blockchain
     val confidenceLevel: Int
+    //BTC node host name
+    val host:String
```
Alexey-N-Chernyshov: Add space for formatting.

PR #223, opened 2018-10-10
Feature/iroha healthcheck

x3medima17:
Now all services wait for Iroha before starting

### DIFF
--- a/build_and_push_nexus.sh
+++ b/build_and_push_nexus.sh
@@ -19,10 +19,10 @@ checkTag $TAG "${tags[@]}"
 rm build/libs/notary-1.0-SNAPSHOT-all.jar || true
 gradle shadowJar
 
-docker build -t nexus.iroha.tech:19002/d3-deploy/eth-relay:$TAG -f eth-relay.dockerfile .
-docker build -t nexus.iroha.tech:19002/d3-deploy/registration:$TAG  -f registration.dockerfile .
-docker build -t nexus.iroha.tech:19002/d3-deploy/notary:$TAG  -f notary.dockerfile .
-docker build -t nexus.iroha.tech:19002/d3-deploy/withdrawal:$TAG  -f withdrawal.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/eth-relay:$TAG -f docker/eth-relay.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/registration:$TAG  -f docker/registration.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/notary:$TAG  -f docker/notary.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/withdrawal:$TAG  -f docker/withdrawal.dockerfile .
 
 
 docker push nexus.iroha.tech:19002/d3-deploy/eth-relay:$TAG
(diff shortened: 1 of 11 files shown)


docker/entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+    echo "Please set IROHA_HOST, IROHA_PORT and CLASS variables"
+    exit
+fi
+
+STATE=0
+
+while [ $STATE -ne 1 ]; do
+    let STATE="$(./grpc_healthcheck ${IROHA_HOST} ${IROHA_PORT})"
+    echo "Waiting for iroha..."
+    sleep 0.1
```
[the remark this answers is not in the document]
x3medima17: That script itself takes more than a second

build_and_push_nexus.sh
```diff
@@ -19,10 +19,10 @@ checkTag $TAG "${tags[@]}"
...
 gradle shadowJar
 
-docker build -t nexus.iroha.tech:19002/d3-deploy/eth-relay:$TAG -f eth-relay.dockerfile .
-docker build -t nexus.iroha.tech:19002/d3-deploy/registration:$TAG  -f registration.dockerfile .
-docker build -t nexus.iroha.tech:19002/d3-deploy/notary:$TAG  -f notary.dockerfile .
-docker build -t nexus.iroha.tech:19002/d3-deploy/withdrawal:$TAG  -f withdrawal.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/eth-relay:$TAG -f docker/eth-relay.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/registration:$TAG  -f docker/registration.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/notary:$TAG  -f docker/notary.dockerfile .
+docker build -t nexus.iroha.tech:19002/d3-deploy/withdrawal:$TAG  -f docker/withdrawal.dockerfile .
```
Warchant: don't use scripts to build docker images. there is an easier way:
https://github.com/soramitsu/BCA-Sora-Unit-Backend/blob/develop/build.gradle#L14
https://github.com/soramitsu/BCA-Sora-Unit-Backend/blob/develop/build.gradle#L149-L155
https://github.com/soramitsu/BCA-Sora-Unit-Backend/blob/develop/Dockerfile
https://github.com/soramitsu/BCA-Sora-Unit-Backend/blob/develop/Jenkinsfile#L21

deploy/docker-compose-full.yml
```diff
@@ -38,6 +38,9 @@ services:
       - ../configs:/opt/notary/configs
     environment:
       - PROFILE
+      - IROHA_HOST=d3-iroha
+      - IROHA_PORT=50051
+      - CLASS=registration.eth.relay.DeployRelayMain
```
Warchant: you don't need to specify main class. just apply 'application' plugin and specify variable `mainClassName` inside gradle.
Then, if you use docker plugin, it will work "as is"

x3medima17: We can't, because our services are not split 
So we have one application but many classnames that we have to run

Warchant: you need to create separate gradle.build for each of projects then.

see https://github.com/soramitsu/BCA-Sora-Middleware/

Warchant: Especially pay attention to https://github.com/soramitsu/BCA-Sora-Middleware/blob/master/settings.gradle

PR #224, opened 2018-10-11
Feature/longevity

Alexey-N-Chernyshov:
# About
Introduce longevity test.
Creates 4 instances of notary and N clients.
Preforms very simple scenario in average load in infinite loop:
- Deposit asset followed by withdraw asset.

PR affects integration helper:
- fix master smart contract lock
- fix configs
- introduce NotaryClient class

Problem with iroha were figured out:
- grpc streaming returns UNKNOWN error after some time (seems uncaught exception in streaming)
- iroha returns empty error response on query (looks like a race condition)

Currently due bugs there is only one client in scenario.

# How to run
`gradle longevityTest`


configs/eth/ethereum_password_local.properties
```diff
@@ -4,3 +4,12 @@ relay-vacuum.credentialsPassword=user
 eth-notary.credentialsPassword=user
 relay-registration.credentialsPassword=user
 eth-registration.credentialsPassword=user
+client0.credentialsPassword=user
```
Alexey-N-Chernyshov: I'm not sure, should this file be committed?

neewy:
I need to know how to run it and what exact commands to run in order to leave it for several hours or days. Can you please attach to PR "How to test" part?

src/main/kotlin/sidechain/iroha/consumer/IrohaNetworkImpl.kt
```diff
@@ -102,8 +102,18 @@ class IrohaNetworkImpl(host: String, port: Int) : IrohaNetwork {
...
 
-        val results = hashes.map { checkTransactionStatus(it) }.filter { it.component1() != null }.map { it.get() }
-        return Result.of { results }
+        return Result.of {
+            hashes.map { checkTransactionStatus(it) }
+                .filter {
+                    it.fold(
+                        { true },
+                        {
+                            logger.warn("Send batch: ", it)
```
dolgopolovwork: Can we add some details in this warning? "Send batch" doesn't sound like something bad happened.

PR #225, opened 2018-10-11
Feature/deploy app

Alexey-N-Chernyshov:
# About
Gradle task to run ethereum predeploy smart contracts. What it does:
- deploy RelayRegistry
- deploy Master
- add all notaries addresses to Master
- lock master

# Known issues
Predeploy uses registry service Ethereum credentials since only owner of RelayRegistry smart contract can add relays.
We should specify how to pass notaries ethereum addresses. Now it is passed via app parameters, but there is mess with ethereum credentials. We should decide how to pass them in a secure and handy way.

# How to run
`gradle runPreDeployEthereum`


src/main/kotlin/deploy/main.kt
```diff
@@ -0,0 +1,38 @@
...
+import sidechain.eth.util.DeployHelper
+
+private val logger = KLogging().logger
+
+/**
+ * Entry point to deploy smart contracts.
+ * [args] should contain the list of notary ethereum addresses
+ */
+fun main(args: Array<String>) {
+    args.forEach { println(it) }
```
x3medima17: Please use logger

PR #226, opened 2018-10-11
Split ContractsTest to master, relay and rr test

luckychess:
Small refactor of contract tests.


Alexey-N-Chernyshov:
I think ContractTestHelper can be part of IntegrationHelperUtil. All contract related code should be moved from IntegrationHelperUtil to ContractTestHelper.

PR #227, opened 2018-10-11
Time from Bitcoin blockchain

dolgopolovwork:
Now tx time is taken directly from blocks. This change leads us to use raw blockchain api instead of wallet api, that does all the dirty work with parsing.


btc-deposit/src/main/kotlin/notary/btc/listener/BitcoinBlockChainListener.kt
```diff
@@ -0,0 +1,51 @@
...
+    private val emitter: ObservableEmitter<SideChainEvent.PrimaryBlockChainEvent>,
+    private val confidenceLevel: Int
+) : BlocksDownloadedEventListener {
+
+    override fun onBlocksDownloaded(peer: Peer?, block: Block?, filteredBlock: FilteredBlock?, blocksLeft: Int) {
+        if (block == null) {
+            logger.warn { "Cannot handle null BTC block" }
+            return
+        } else if (block.time.time < System.currentTimeMillis() - DAY_MILLIS) {
+            //We cannot handle too old blocks due to Iroha time restrictions.
```
grimadas: Should we do something in this case?

dolgopolovwork: We cannot do anything in this case. At least at d3 side.

PR #239, opened 2018-10-15
Add docker-machine to ansible

x3medima17:
Now we can use docker machine to deploy D3 on.


deploy/ansible/inventory.list
```diff
@@ -1,5 +1,19 @@
...
+node0 ansible_host=212.47.240.224  ansible_user=root notary_key=notary1
+node1 ansible_host=51.15.231.25  ansible_user=root  notary_key=notary2
+node2 ansible_host=163.172.187.40  ansible_user=root  notary_key=notary3
+
+[ubuntu:vars]
+ansible_ssh_private_key_file=/Users/dumitru/.ssh/notary
+
+
+[docker]
+vm0 ansible_host=192.168.99.101  ansible_ssh_private_key_file=/Users/dumitru/.docker/machine/machines/node0/id_rsa
```
bakhtin: What are these machines with private addresses?

deploy/ansible/main.yml
```diff
@@ -1,13 +1,24 @@
...
 
-    roles:
-      - { role: iroha }
-      - { role: docker }
-      - { role: notary }
+- hosts: docker
+  gather_facts: yes
+  pre_tasks:
+    - name: install python 2
+      raw: tce-load -wi python
```
bakhtin: Why do we use that `tce-load` instead of `apt` package manager? Is that some kind of a custom Linux distro that you assume `tce-load` is pre-installed on it?

deploy/ansible/main.yml
```diff
@@ -1,13 +1,24 @@
...
+      become: false
+
+    - name: Install pip
```
bakhtin: Pip can be installed inside main `tasks` file. This package is not essential for Ansible operations. And, please, install it via `apt` Ansible module.

deploy/ansible/roles/docker/tasks/clean_docker.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: remove all docker containers
```
bakhtin: You should not assume it is safe to clean up all the images and containers on the system. Other containers may also present on the same system.

deploy/ansible/roles/docker/tasks/main.yml
```diff
@@ -1,3 +1,27 @@
...
+- name: install docker python modules
+  pip:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - docker
+    - docker-compose
+
+
+- include_tasks: install_docker.yml
```
bakhtin: You can safely omit the check for Linux distribution. In this and following playbooks assume it is Ubuntu Linux as we use it as the primary distro. We also gradually remove these checks in our playbooks. YAML files in `tasks` can also be simply named `main.yml` (not `ubuntu.yaml` included into `main.yml` as previously)

deploy/ansible/inventory.list
```diff
@@ -1,5 +1,19 @@
...
-node0 ansible_host=212.47.240.224  ansible_user=root
-node1 ansible_host=51.15.231.25  ansible_user=root
-node2 ansible_host=163.172.187.40  ansible_user=root
-node3 ansible_host=163.172.146.134 ansible_user=root
+node0 ansible_host=212.47.240.224  ansible_user=root notary_key=notary1
+node1 ansible_host=51.15.231.25  ansible_user=root  notary_key=notary2
+node2 ansible_host=163.172.187.40  ansible_user=root  notary_key=notary3
+
+[ubuntu:vars]
+ansible_ssh_private_key_file=/Users/dumitru/.ssh/notary
```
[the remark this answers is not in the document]
x3medima17: They have different keys

deploy/ansible/main.yml
x3medima17: Yes, docker machine uses boot2docker, it does not have apt

deploy/ansible/main.yml
```diff
@@ -1,13 +1,24 @@
...
+      raw: curl https://bootstrap.pypa.io/get-pip.py | python -
+
+- hosts: all
+  gather_facts: yes
+  roles:
+    - { role: docker }
+#      - { role: iroha }
```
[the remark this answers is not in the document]
x3medima17: At this moment yes, when the trunk is ready, I will uncomment it

deploy/ansible/inventory.list
x3medima17: Just some machines created locally.

deploy/ansible/roles/docker/tasks/main.yml
x3medima17: Sometimes I use distro, other than ubuntu

x3medima17:
TODO:
Add readme(why docker and ubuntu), blank keypath

PR #242, opened 2018-10-16
Move some logic from IHU to ContractTestHelper

luckychess:
Just a small refactoring.


notary-integration-test/src/main/kotlin/integration/helper/IntegrationHelperUtil.kt
```diff
@@ -87,7 +79,7 @@ class IntegrationHelperUtil {
     val ethRegistrationConfig by lazy { configHelper.createEthRegistrationConfig(testConfig.ethereum) }
 
     /** Ethereum utils */
-    private val deployHelper by lazy { DeployHelper(testConfig.ethereum, ethPasswordConfig) }
+    private val contractTestHelper = ContractTestHelper()
```
Alexey-N-Chernyshov: mb make it lazy?

PR #252, opened 2018-10-22
Bitcoin wallet is ignored

dolgopolovwork:
We had a problem with bitcoin wallet file being tracked by git even with well configured `.gitignore`. Git tracks wallet file because it was commited in the past. We cannot remove this file, because it's needed to run integration tests. The simpliest workaround is creating ignored temporary copy of wallet file every time we run bitcoin tests.

### DIFF
--- a/notary-integration-test/src/main/kotlin/integration/helper/ConfigHelper.kt
+++ b/notary-integration-test/src/main/kotlin/integration/helper/ConfigHelper.kt
@@ -12,6 +12,9 @@ import registration.eth.relay.RelayRegistrationConfig
 import token.ERC20TokenRegistrationConfig
 import vacuum.RelayVacuumConfig
 import withdrawalservice.WithdrawalServiceConfig
+import java.io.File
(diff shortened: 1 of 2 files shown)


notary-integration-test/src/main/kotlin/integration/helper/ConfigHelper.kt
```diff
@@ -71,11 +74,25 @@ class ConfigHelper(
...
     }
 
+    /*
+        Creates temporary copy of wallet file just for testing
+     */
+    fun createTempWalletFile(btcWalletPath: String): String {
+        val newWalletFilePath = btcWalletPath.replaceFirst(".wallet", "-test.wallet")
+        val walletFile = File(newWalletFilePath)
+        if (!walletFile.exists()) {
+            val src = FileInputStream(btcWalletPath).channel
```
dolgopolovwork: Seems like I forgot to close these resources.

PR #265, opened 2018-11-07
Storing public keys for multisignature addresses

dolgopolovwork:
Now we store all public keys that were used in multisignature address creation process. Keys are stored in JSON.

### DIFF
--- a/btc-registration/src/main/kotlin/registration/btc/BtcRegistrationStrategyImpl.kt
+++ b/btc-registration/src/main/kotlin/registration/btc/BtcRegistrationStrategyImpl.kt
@@ -5,10 +5,12 @@ import com.github.kittinunf.result.fanout
 import com.github.kittinunf.result.flatMap
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
+import provider.btc.address.AddressInfo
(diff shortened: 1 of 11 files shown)


btc-registration/src/main/kotlin/registration/btc/BtcRegistrationStrategyImpl.kt
```diff
@@ -30,8 +32,19 @@ class BtcRegistrationStrategyImpl(
             .flatMap { (addresses, takenAddresses) ->
                 try {
                     //It fetches all BTC addresses and takes one that was not registered
-                    val freeAddress = addresses.keys.first { btcAddress -> !takenAddresses.containsKey(btcAddress) }
-                    irohaAccountCreator.create(freeAddress, whitelist.toString().trim('[').trim(']'), name, pubkey)
+                    val freeAddress =
+                        addresses.first { btcAddress -> !takenAddresses.any { takenAddress -> takenAddress.address == btcAddress.address } }
```
grimadas: Possibly not thread safe, leave todo

dolgopolovwork: This is definitely not thread safe and will cause a lot of troubles. Thank you.

PR #267, opened 2018-11-09
Bitcoin white list

dolgopolovwork:
Bitcoin whitelist implementation. Works the same way as in Ethereum services. A whitelist is stored in client account details and checked every time withdrawal event happens.

### DIFF
--- a/src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
+++ b/src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
@@ -3,7 +3,6 @@ package notary.endpoint.eth
 import com.github.kittinunf.result.Result
 import com.github.kittinunf.result.fanout
 import com.github.kittinunf.result.flatMap
-import com.github.kittinunf.result.map
 import config.EthereumConfig
 import config.EthereumPasswords
 import iroha.protocol.TransactionOuterClass.Transaction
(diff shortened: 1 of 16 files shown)


src/main/kotlin/provider/WhiteListProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class WhiteListProvider(
+    private val whiteListSetterAccount: String,
+    private val credential: IrohaCredential,
+    private val irohaNetwork: IrohaNetwork,
+    private val whiteListKey: String
+) {
+
+    /**
+     * Check if [srcAccountId] has withdrawal [address] in whitelist
```
grimadas: Please add comment about behaviour in case when whitelist is empty

src/main/kotlin/registration/IrohaAccountCreator.kt
```diff
@@ -20,26 +20,7 @@ class IrohaAccountCreator(
...
-    fun create(
-        currencyAddress: String,
-        whitelist: String,
-        userName: String,
-        pubkey: String
-    ): Result<String, Exception> {
-        return create(currencyAddress, whitelist, userName, pubkey) { "$userName@$CLIENT_DOMAIN" }
-    }
-
+    //TODO update comments
```
grimadas: This is unclear

src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
```diff
@@ -27,21 +27,26 @@ class NotaryException(reason: String) : Exception(reason)
...
 
-    val relayProvider = EthRelayProviderIrohaImpl(
+    private val relayProvider = EthRelayProviderIrohaImpl(
         irohaNetwork,
         credential,
         credential.accountId,
         notaryConfig.registrationServiceIrohaAccount
     )
 
+    private val whiteListProvider = WhiteListProvider(
```
grimadas: What about abstract class whiteListProvider and two descendants: EthWhitelist and BtcWhitelist? Maybe this way it will be a better design

src/main/kotlin/registration/IrohaAccountCreator.kt
dolgopolovwork: my bad

src/main/kotlin/notary/endpoint/eth/EthRefundStrategyImpl.kt
dolgopolovwork: This is a quite good idea. Yes, I will do this.

PR #270, opened 2018-11-14
Bitcoin fee computation

dolgopolovwork:
Now bitcoin fee depends on tx size.

### DIFF
--- a/btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
+++ b/btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
@@ -14,8 +14,14 @@ import org.springframework.stereotype.Component
 import provider.btc.address.BtcRegisteredAddressesProvider
 import provider.btc.network.BtcNetworkConfigProvider
 
-//TODO develop more sophisticated mechanism to compute fee
-private const val MIN_FEE = 1000
+//TODO make it configurable
+//Fee per byte rate
+private const val FEE_RATE = 10
+//Only two ouputs are used: destination and change
+private const val OUTPUTS = 2
+
+private const val BYTES_PER_INPUT = 180
+private const val BYTES_PER_OUTPUT = 34
 
 /*
     Helper class that is used to collect inputs, outputs and etc
@@ -48,7 +54,7 @@ class TransactionHelper(
             Coin.valueOf(amount),
             Address.fromBase58(btcNetworkConfigProvider.getConfig(), destinationAddress)
         )
-        val change = totalAmount - amount - MIN_FEE
+        val change = totalAmount - amount - getTxFee(transaction.inputs.size, OUTPUTS)
         //TODO create change address creation mechanism
         transaction.addOutput(Coin.valueOf(change), changeAddress)
     }
@@ -68,34 +74,7 @@ class TransactionHelper(
         confidenceLevel: Int
     ): Result<List<TransactionOutput>, Exception> {
         return Result.of {
-            //Only confirmed transactions must be used
-            val unspents =
-                ArrayList<TransactionOutput>(wallet.unspents.filter { unspent ->
-                    unspent.parentTransactionDepthInBlocks >= confidenceLevel && !usedOutputs.contains(
-                        unspent
-                    )
-                })
-            /*
-            Wallet stores unspents in a HashSet, while the order of HashSet depends on several factors: current array size and etc.
-            This may lead different notary nodes to pick different transactions.
-            This is why we order transactions manually, essentially reducing the probability of
-            different nodes to pick different transactions*/
-            unspents.sortWith(Comparator { output1, output2 -> output1.hashCode().compareTo(output2.hashCode()) })
-            val usedUnspents = ArrayList<TransactionOutput>()
-            val totalAmount = amount + MIN_FEE
-            var collectedAmount = 0L
-            //We use registered clients outputs only
-            unspents.filter { unspent -> isAvailableOutput(availableAddresses, unspent) }.forEach { unspent ->
-                if (collectedAmount >= totalAmount) {
-                    return@forEach
-                }
-                collectedAmount += unspent.value.value
-                usedUnspents.add(unspent)
-            }
-            if (collectedAmount < totalAmount) {
-                throw IllegalStateException("Cannot get enough BTC amount(required $amount, collected $collectedAmount) using current unspent tx collection")
-            }
-            usedUnspents
+            collectUnspentsRec(availableAddresses, wallet, amount, confidenceLevel, ArrayList())
         }
     }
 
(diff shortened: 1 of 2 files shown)


btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
```diff
@@ -14,8 +14,14 @@ import org.springframework.stereotype.Component
 import provider.btc.address.BtcRegisteredAddressesProvider
 import provider.btc.network.BtcNetworkConfigProvider
 
-//TODO develop more sophisticated mechanism to compute fee
-private const val MIN_FEE = 1000
+//TODO make it configurable
+//Fee per byte rate
```
grimadas: Add: Fee rate per byte in SAT

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
```diff
@@ -126,15 +105,97 @@ class TransactionHelper(
...
+                        //Cannot use unspents from another level of recursion
+                        && !recursivelyCollectedUnspents.contains(unspent)
+            })
+        if (unspents.isEmpty()) {
+            throw IllegalStateException("Out of unspents")
+        }
+        /*
+        Wallet stores unspents in a HashSet. Order of a HashSet depends on several factors: current array size and etc.
+        This may lead different notary nodes to pick different transactions.
+        This is why we order transactions manually, essentially reducing the probability of
```
grimadas: You are reducing the probability, but it will be still not zero?

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
```diff
@@ -126,15 +105,97 @@ class TransactionHelper(
...
+                return@forEach
+            }
+            collectedAmount += unspent.value.value
+            recursivelyCollectedUnspents.add(unspent)
+        }
+        if (collectedAmount < amount) {
+            throw IllegalStateException("Cannot get enough BTC amount(required $amount, collected $collectedAmount) using current unspent tx collection")
+        }
+        //Check if able to pay fee
+        val amountAndFee = amount + getTxFee(recursivelyCollectedUnspents.size, OUTPUTS)
```
grimadas: But this will add recursively fee each time, right?

notary-integration-test/src/integration-test/kotlin/integration/btc/BtcWithdrawalIntegrationTest.kt
```diff
@@ -108,13 +108,19 @@ class BtcWithdrawalIntegrationTest {
...
     fun setUp() {
+        val blockStorageFolder = File(btcWithdrawalConfig.bitcoin.blockStoragePath)
+        //Clear bitcoin blockchain folder
+        blockStorageFolder.deleteRecursively()
+        //Recreate folder
+        blockStorageFolder.mkdirs()
         integrationHelper.generateBtcBlocks()
         //TODO make it faster
-        integrationHelper.preGenBtcAddresses(btcWithdrawalConfig.bitcoin.walletPath, 14).failure { ex -> throw ex }
+        integrationHelper.preGenBtcAddresses(btcWithdrawalConfig.bitcoin.walletPath, 18).failure { ex -> throw ex }
```
grimadas: Why 18?

notary-integration-test/src/integration-test/kotlin/integration/btc/BtcWithdrawalIntegrationTest.kt
```diff
@@ -162,6 +168,73 @@ class BtcWithdrawalIntegrationTest {
...
+            btcAsset,
+            btcAddressDest,
+            amount.toString()
+        )
+        Thread.sleep(WITHDRAWAL_WAIT_MILLIS)
+        assertEquals(initTxCount, createdTransactions.size)
+        transactionHelper.addToBlackList(btcAddressSrc)
+        transactionHelper.addToBlackList(btcAddressDest)
+    }
+
```
grimadas: Also, add test case when spending is possible, but control the fee to be in some limit

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
dolgopolovwork: I can't say for sure. There are too many factors to take into account. For example, we may be connected to extremely misbehaving Bitcoin node with completely different blockchain. So, yes. It's not zero.

notary-integration-test/src/integration-test/kotlin/integration/btc/BtcWithdrawalIntegrationTest.kt
dolgopolovwork: Can you please add some details? I cannot understand this test case.

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
grimadas: The peergroup node is our, so I doubt that it will be misbehaving

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
grimadas: Better use const amount and dynamic fee variable, otherwise this recursion will not terminate optimally

notary-integration-test/src/integration-test/kotlin/integration/btc/BtcWithdrawalIntegrationTest.kt
grimadas: Maybe some unit cases in which you expect to take specific outputs

notary-integration-test/src/integration-test/kotlin/integration/btc/BtcWithdrawalIntegrationTest.kt
grimadas: Put a constant value with explanation

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/TransactionHelper.kt
```diff
@@ -111,6 +111,7 @@ class TransactionHelper(
      * @param availableAddresses - set of addresses which transactions will be available to spend
      * @param wallet - current wallet. Used to fetch unspents
      * @param amount - amount of SAT to spend
+     * @param fee - tx fee that depends on inputs and outputs. Initial value is zero.
```
grimadas: Can we use default value then?

dolgopolovwork: No

dolgopolovwork: Hmm. Yes, we can do it. Thank you.

grimadas: So maybe we should split the function, one will be for internal use(this one), and other will call internal, because this function will kind of calculate optimal and necessary fee amount, than I should I put zero as a param

grimadas: Already solved

PR #287, opened 2018-11-27
Feature/tests timeouts

Mingela:
### Description of the Change
Added awkward integration tests timeouts. 

Details: 
https://github.com/junit-team/junit5/issues/80#issuecomment-391313036 
https://blog.philipphauer.de/best-practices-unit-testing-kotlin/

On the one hand it is generally bad idea to use JUnit4 for Kotlin
On the other hand there is no convenient way to implement timeouts in JUnit5 yet

### DIFF
--- a/notary-integration-test/src/integration-test/kotlin/integration/btc/BtcNotaryIntegrationTest.kt
+++ b/notary-integration-test/src/integration-test/kotlin/integration/btc/BtcNotaryIntegrationTest.kt
@@ -2,7 +2,6 @@ package integration.btc
 
 import com.github.kittinunf.result.failure
 import handler.btc.NewBtcClientRegistrationHandler
-import integration.helper.ConfigHelper
 import integration.helper.IntegrationHelperUtil
 import integration.helper.btcAsset
 import listener.btc.NewBtcClientRegistrationListener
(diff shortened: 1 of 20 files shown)


Mingela:
Let me know if there is no sense in a such wrapping for some tests

configs/test_local.properties
```diff
@@ -25,3 +25,5 @@ test.ethereum.gasLimit=6000000
 # --------- Bitcoin ---------
 test.bitcoin.walletPath=deploy/bitcoin/test-btc.wallet
 test.bitcoin.blockStoragePath=deploy/bitcoin/blocks
+# --------- Timeout (minutes) ---------
+test.testTimeout=5
```
dolgopolovwork: I think we can hardcode this value.

notary-integration-test/src/integration-test/kotlin/integration/btc/BtcNotaryIntegrationTest.kt
```diff
@@ -85,22 +85,24 @@ class BtcNotaryIntegrationTest {
...
-        val btcAmount = 1
-        integrationHelper.sendBtc(btcAddress, btcAmount)
-        Thread.sleep(20_000)
-        val newBalance = integrationHelper.getIrohaAccountBalance(testClient, btcAsset)
-        assertEquals(
-            BigDecimal(initialBalance).add(integrationHelper.btcToSat(btcAmount).toBigDecimal()).toString(),
-            newBalance
-        )
-        assertTrue(btcNotaryInitialization.isWatchedAddress(btcAddress))
+        assertTimeoutPreemptively(timeoutDuration) {
```
dolgopolovwork: I don't think we need `assertTimeoutPreemptively` in every test. Can we use this construction in tests known to be long-runing only?

configs/test_local.properties
Alexey-N-Chernyshov: Agree.

Mingela: Done as a constant value In ConfigHelper

Alexey-N-Chernyshov:
Please, don't use force push for your branches, since it is impossible to view changes from the last review.

PR #288, opened 2018-11-27
Bitcoin monitoring

dolgopolovwork:
### Description of the Change
Spring based JMX monitoring was added. You can get this information using jconsole or any other JMX client.

### DIFF
--- a/btc-withdrawal/src/main/kotlin/withdrawal/btc/handler/WithdrawalTransferEventHandler.kt
+++ b/btc-withdrawal/src/main/kotlin/withdrawal/btc/handler/WithdrawalTransferEventHandler.kt
@@ -4,6 +4,7 @@ import com.github.kittinunf.result.failure
 import com.github.kittinunf.result.map
 import helper.address.isValidBtcAddress
 import iroha.protocol.Commands
+import monitoring.Monitoring
 import mu.KLogging
 import org.bitcoinj.core.Transaction
 import org.bitcoinj.wallet.Wallet
@@ -11,6 +12,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 import withdrawal.btc.config.BtcWithdrawalConfig
 import withdrawal.btc.provider.BtcWhiteListProvider
+import withdrawal.btc.statistics.WithdrawalStatistics
 import withdrawal.btc.transaction.SignCollector
 import withdrawal.btc.transaction.TransactionCreator
 import withdrawal.btc.transaction.UnsignedTransactions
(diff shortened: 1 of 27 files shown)


btc-withdrawal/src/main/kotlin/withdrawal/btc/statistics/WithdrawalStatistics.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Data class that holds short statistics about withdrawal service
+ */
+data class WithdrawalStatistics(
+    val totalTransfers: AtomicInteger,
+    val failedTransfers: AtomicInteger,
+    val succeededTransfers: AtomicInteger
+) {
+    fun incTotalTransfers() = totalTransfers.incrementAndGet()
```
Mingela: Do we need to keep a total value which is always a sum of failed and succeeded? Or not?

src/main/kotlin/util/StringExt.kt
```diff
@@ -53,3 +57,7 @@ fun String.Companion.unHex(s: String): ByteArray {
 fun String.Companion.irohaEscape(text: String): String {
     return text.replace("\"", IROHA_FRIENDLY_QUOTE)
 }
+
+//TODO can we get rid of klaxon and moshi? Gson is much easier thing to use.
```
Mingela: I propose to create a task in backlog for that

btc-withdrawal/src/main/kotlin/withdrawal/btc/handler/WithdrawalTransferEventHandler.kt
```diff
@@ -64,6 +70,7 @@ class WithdrawalTransferEventHandler(
                     logger.warn { "Cannot withdraw to $destinationAddress, because it's not in ${transferCommand.srcAccountId} whitelist" }
                 }
             }, { ex ->
+                withdrawalStatistics.incFailedTransfers()
```
Mingela: Didn't you forget to increment succeeded ones?

btc-withdrawal/src/main/kotlin/withdrawal/btc/statistics/WithdrawalStatistics.kt
dolgopolovwork: In my opinion, we need this. For example, I don't want to do any math. I just want to see numbers.

src/main/kotlin/util/StringExt.kt
Mingela: IMO Unification of the tools used in a project is always a good proposal. We can define a potential task without specifying the concrete library, but list of possible candidates and vote afterwards after discussion who is better and why.

btc-withdrawal/src/main/kotlin/withdrawal/btc/statistics/WithdrawalStatistics.kt
Mingela: Time vs Space holy war confirmed :) So do we need to estimate how often we update the total value and determine which overhead is more significant?

grimadas: I don't think the overhead is big, so we can leave the value as it is

src/main/kotlin/util/StringExt.kt
Alexey-N-Chernyshov: I think we can use https://github.com/Kotlin/kotlinx.serialization. It is shipped with Kotlin since 1.3. Also I've added a task to migrate.

PR #289, opened 2018-11-28
Bitcoin withdrawal-registration sync

dolgopolovwork:
### Description of the Change
Registration and withdrawal services are well synchronized now. A newly registered address will be handled by withdrawal component.

### DIFF
--- a/btc/src/main/kotlin/handler/btc/NewBtcClientRegistrationHandler.kt
+++ b/btc/src/main/kotlin/handler/btc/NewBtcClientRegistrationHandler.kt
@@ -1,0 +1,49 @@
+package handler.btc
+
+import iroha.protocol.Commands
+import mu.KLogging
+import org.bitcoinj.core.Address
+import org.bitcoinj.wallet.Wallet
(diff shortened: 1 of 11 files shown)


btc/src/main/kotlin/handler/btc/NewBtcClientRegistrationHandler.kt
```diff
@@ -0,0 +1,49 @@
...
+) {
+    /**
+     * Handles newly registered Bitcoin addresses and adds addresses to current wallet object
+     */
+    fun handleNewClientCommand(command: Commands.Command, wallet: Wallet) {
+        if (!isNewClientWasRegistered(command)) {
+            return
+        }
+        val address = Address.fromBase58(
+            btcNetworkConfigProvider.getConfig(),
```
Mingela: Are these configs changeable in runtime? If not, can we just `getConfig()` during initialization and reuse only that value?

dolgopolovwork: The value returned by `getConfig()` is a singleton.

PR #295, opened 2018-11-28
separate integration tests on different modules by sidechains

Mingela:
### Description of the Change
Integration tests are now divided into corresponding modules
Also updated Jenkinsfile


dolgopolovwork:
Looks cool

Mingela:
Also tried to execute tests in parallel with JUnut 5.3.2 but many of test assertions failed. We may need to discuss what could be processed in parallel etc.

PR #299, opened 2018-12-02
function for adding and removing peers

Hirama:
the constructor accepts arrays of initial peers
function for adding and removing peers


Alexey-N-Chernyshov:
Do we need locking master anymore? Can you add a test to show how it works now, please?

PR #302, opened 2018-12-04
Refactoring: Eth modules

Mingela:
### Description of the Change
Moved ethereum functionality to separate modules


eth-erc20-registration/src/main/kotlin/token/ERC20TokenRegistration.kt
```diff
@@ -41,7 +40,7 @@ class ERC20TokenRegistration(
                     Result.of { logger.warn { "No ERC20 tokens to register" } }
                 } else {
                     EthTokensProviderImpl.logger.info { "ERC20 tokens to register $tokensToRegister" }
```
dolgopolovwork: It's not your fault, but I can see a little problem here. Logger was taken from another file by mistake.

eth/src/main/java/contract/RelayRegistry.java
```diff
@@ -86,7 +86,7 @@ protected RelayRegistry(String contractAddress, Web3j web3j, TransactionManager
```
dolgopolovwork: Can we modify this file? I know that it was generated.

eth-erc20-registration/src/main/kotlin/token/ERC20TokenRegistration.kt
Mingela: Thank you for detecting this problem

eth/src/main/java/contract/RelayRegistry.java
Mingela: Could you be more specific on that? What exact changes would you like to apply?

dolgopolovwork: I think you should ask @Alexey-N-Chernyshov .

build.gradle
```diff
@@ -169,9 +169,9 @@ allprojects {
     test {
         setWorkingDir("$rootDir/")
         useJUnitPlatform()
-        systemProperty 'java.library.path', Paths.get(projectDir.toString(), "iroha_bindings", getCurrentOs()).toString()
+        systemProperty 'java.library.path', Paths.get(projectDir.getParent().toString(), "iroha_bindings", getCurrentOs()).toString()
         println "Runtime Iroha library path:${systemProperties}"
```
dolgopolovwork: Can we remove all dependencies related to Ethereum? I don't need them in btc modules.

Mingela: I have moved the root project dependencies further to subprojects (modules)

eth/src/main/java/contract/BasicCoin.java
```diff
@@ -91,16 +91,16 @@ protected BasicCoin(String contractAddress, Web3j web3j, TransactionManager tran
     public RemoteCall<TransactionReceipt> approve(String _spender, BigInteger _value) {
         final Function function = new Function(
                 FUNC_APPROVE, 
-                Arrays.<Type>asList(new org.web3j.abi.datatypes.Address(_spender), 
-                new org.web3j.abi.datatypes.generated.Uint256(_value)), 
+                Arrays.<Type>asList(new Address(_spender),
```
Alexey-N-Chernyshov: This file is autogenerated code, do not modify this, please.

eth/src/main/java/contract/Master.java
```diff
@@ -85,8 +85,8 @@ protected Master(String contractAddress, Web3j web3j, TransactionManager transac
     public RemoteCall<TransactionReceipt> addPeers(List<String> newAddresses) {
         final Function function = new Function(
                 FUNC_ADDPEERS, 
-                Arrays.<Type>asList(new org.web3j.abi.datatypes.DynamicArray<org.web3j.abi.datatypes.Address>(
-                        org.web3j.abi.Utils.typeMap(newAddresses, org.web3j.abi.datatypes.Address.class))), 
+                Arrays.<Type>asList(new DynamicArray<Address>(
```
Alexey-N-Chernyshov: sap

eth/src/main/java/contract/Relay.java
```diff
@@ -80,33 +80,33 @@ protected Relay(String contractAddress, Web3j web3j, TransactionManager transact
     public RemoteCall<TransactionReceipt> sendToMaster(String tokenAddress) {
         final Function function = new Function(
                 FUNC_SENDTOMASTER, 
-                Arrays.<Type>asList(new org.web3j.abi.datatypes.Address(tokenAddress)), 
+                Arrays.<Type>asList(new Address(tokenAddress)),
```
Alexey-N-Chernyshov: sap

eth/src/main/java/contract/RelayRegistry.java
Alexey-N-Chernyshov: Anton is correct. All java/contract/*.java files are auto-generated code that shouldn't be modified.

eth/src/main/java/contract/BasicCoin.java
Mingela: Sure I will rollback the changes. Probably it was done by IDE automatically.

build.gradle
```diff
@@ -44,9 +44,6 @@ allprojects {
 }
```
dolgopolovwork: Can we also move `run` tasks to dedicated modules?

eth/src/main/java/contract/RelayRegistry.java
Mingela: Done

build.gradle
Mingela: It can be done, but it will lead to either moving all related configs to modules or addition paths processing (in every file/method related to configs, keys etc) since we need to work with 'parent' directory from a module.
So the easiest (and probably right) way is to have gradle tasks on the same level with the related materials. (Both configs and tasks at the root OR both configs and tasks in the corresponding module).
Should we do such refactoring and if yes what is the exact strategy? @Alexey-N-Chernyshov @dolgopolovwork

dolgopolovwork: Please take a look at `btc` services. We can do it with ease.

Mingela: Thank you for the advice
I hope I did it correctly

PR #304, opened 2018-12-04
Multiple nodes

dolgopolovwork:
### Description of the Change
In a current implementation, we can be connected to one full node only. But we may increase system stability by connecting to multiple full nodes.

### DIFF
--- a/btc/src/main/kotlin/helper/network/BitcoinNetworkHelper.kt
+++ b/btc/src/main/kotlin/helper/network/BitcoinNetworkHelper.kt
@@ -15,9 +15,12 @@ private val logger = KLogging().logger
 /**
  * Starts bitcoin blockchain downloading process
  */
-fun startChainDownload(peerGroup: PeerGroup, host: String) {
+fun startChainDownload(peerGroup: PeerGroup, hosts: List<String>) {
     logger.info { "Start bitcoin blockchain download" }
-    peerGroup.addAddress(InetAddress.getByName(host))
+    hosts.forEach { host ->
+        peerGroup.addAddress(InetAddress.getByName(host))
+        logger.info { "$host was added to peer group" }
+    }
     peerGroup.startAsync()
     peerGroup.downloadBlockChain()
 }
--- a/src/main/kotlin/config/Configs.kt
+++ b/src/main/kotlin/config/Configs.kt
@@ -61,8 +61,12 @@ interface BitcoinConfig {
     val blockStoragePath: String
     //Depth of transactions in BTC blockchain
     val confidenceLevel: Int
-    //BTC node host name
-    val host: String
+    //BTC node hosts
+    val hosts: String
+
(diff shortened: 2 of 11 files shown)


btc-deposit/src/main/kotlin/notary/btc/config/BtcNotaryAppConfiguration.kt
```diff
@@ -54,4 +54,4 @@ class BtcNotaryAppConfiguration {
         notaryConfig.iroha.port,
         notaryCredential
     )
-}
+}
```
Mingela: Just why :)

src/main/kotlin/config/Configs.kt
```diff
@@ -61,8 +61,12 @@ interface BitcoinConfig {
     val blockStoragePath: String
     //Depth of transactions in BTC blockchain
     val confidenceLevel: Int
-    //BTC node host name
-    val host: String
+    //BTC node hosts
+    val hosts: String
+
+    companion object {
+        fun extractHosts(bitcoinConfig: BitcoinConfig) = bitcoinConfig.hosts.replace(" ", "").split(",")
```
Mingela: I think it is better to use a different approach. I mean something closer to `commons-configurations` implementation should be used so that we could have `List<String>` or `String[]`  in our config class and parse the value into the required type at the config loading phase.

btc/src/main/kotlin/helper/network/BitcoinNetworkHelper.kt
```diff
@@ -15,9 +15,12 @@ private val logger = KLogging().logger
 /**
  * Starts bitcoin blockchain downloading process
  */
-fun startChainDownload(peerGroup: PeerGroup, host: String) {
+fun startChainDownload(peerGroup: PeerGroup, hosts: List<String>) {
     logger.info { "Start bitcoin blockchain download" }
-    peerGroup.addAddress(InetAddress.getByName(host))
+    hosts.forEach { host ->
```
Mingela: Should we foresee something like `hosts.filter { host -> !host.isEmpty() }.forEach {` ?

dolgopolovwork: I don't think so. Host will be validated by `InetAddress.getByName()`

btc-deposit/src/main/kotlin/notary/btc/config/BtcNotaryAppConfiguration.kt
dolgopolovwork: My bad

src/main/kotlin/config/Configs.kt
dolgopolovwork: cfg4k is not very flexible to do this. I think we should move configurations files to JSON/YAML format or try another configuration framework.

PR #311, opened 2018-12-12
Creating block storage folder at init phase

dolgopolovwork:
### Description of the Change
No need to create a block storage folder manually. It will be created by itself.

### DIFF
--- a/notary-commons/src/main/kotlin/util/FileUtil.kt
+++ b/notary-commons/src/main/kotlin/util/FileUtil.kt
@@ -1,0 +1,16 @@
+package util
+
+import java.io.File
+import java.io.IOException
+
+/**
+ * Creates folder with all parent sub-folders.
+ * @param folderPath - path of folder to create
+ * @throws IOException, if it's impossible to create folder
+ */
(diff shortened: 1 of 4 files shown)


notary-commons/src/main/kotlin/util/FileUtil.kt
```diff
@@ -0,0 +1,16 @@
...
+import java.io.IOException
+
+/**
+ * Creates folder with all parent sub-folders.
+ * @param folderPath - path of folder to create
+ * @throws IOException, if it's impossible to create folder
+ */
+fun createFolderIfDoesntExist(folderPath: String) {
+    val folder = File(folderPath)
+    if (!folder.exists() && !File(folderPath).mkdirs()) {
```
Mingela: `mkdirs()` `Creates the directory named by this abstract pathname, including any necessary but nonexistent parent directories. `
Can't we just have `if(!folder.mkdirs())` ?

dolgopolovwork: No, because this function returs false, if folder exists.

PR #315, opened 2018-12-14
Boot repackage was disabled

dolgopolovwork:
### Description of the Change
Spring Boot plugin adds a task called `bootRepackage` that packages spring project into an executable jar file. `btc` is a library project with no main class, meaning that it cannot be executed directly. That's why we don't need this task in this project.


x3medima17:
It still does not work with `gradle build`
Maybe it makes sense to fix the whole build in one PR?

dolgopolovwork:
> It still does not work with `gradle build`
> Maybe it makes sense to fix the whole build in one PR?
This `pr` is not strictly related to build problems. We just have no need in `bootRepackage` in this case.

PR #321, opened 2018-12-19
Bitcoin rollback

dolgopolovwork:
### Description of the Change
Now we can rollback Bitcoin withdrawal transactions in a multisig fashion. The following cases are covered:
1) Client sent 'dusty' money
2) Client sent money to Bitcoin address that is not in a whitelist
3) Client sent money to an invalid Bitcoin address
4) Not enough Bitcoins to create a transaction


dolgopolovwork:
>Are there some tests for as many rollback cases as possible
Yep. I test that money is back in case of failures described above.

PR #324, opened 2018-12-24
Ignoring failed Iroha account creation command

dolgopolovwork:
### Description of the Bug
Two nodes cannot perform a withdrawal transaction simultaneously, because every node will try to create "tx signature storage" account. Only one account creation command will be successfully committed. So we have to ignore account creation command exceptions.

### DIFF
--- a/btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/SignCollector.kt
+++ b/btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/SignCollector.kt
@@ -1,7 +1,6 @@
 package withdrawal.btc.transaction
 
 import com.github.kittinunf.result.Result
-import com.github.kittinunf.result.failure
 import com.github.kittinunf.result.flatMap
 import com.github.kittinunf.result.map
 import com.squareup.moshi.Moshi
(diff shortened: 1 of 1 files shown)


btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/SignCollector.kt
```diff
@@ -62,13 +61,13 @@ class SignCollector(
             logger.info { "Tx ${tx.hashAsString} signatures to add in Iroha $signedInputs" }
             val shortTxHash = shortTxHash(tx)
             val createAccountTx = IrohaConverterImpl().convert(createSignCollectionAccountTx(shortTxHash))
+            /**
+             * We create a dedicated account everytime a withdrawal transaction occurred.
```
grimadas: *on every withdrawal event

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/SignCollector.kt
```diff
@@ -62,13 +61,13 @@ class SignCollector(
...
+             * We need this account to store transaction signatures from all the nodes.
+             * Every node will try to create an account, but only one creation will be succeeded.
```
grimadas: *will succeed

btc-withdrawal/src/main/kotlin/withdrawal/btc/transaction/SignCollector.kt
```diff
@@ -62,13 +61,13 @@ class SignCollector(
...
+             * So it's ok to fail the following Iroha command.
```
grimadas: better just: The following Iroha command can fail

PR #325, opened 2018-12-24
Bitcoin multisig address safe time source

dolgopolovwork:
### Description of the Bug
We use Iroha multisig transactions to create Bitcoin addresses. A previous source of time was a node time, meaning that multisig will never be succeeded because it will have different hash value on different nodes. Now we set time at caller level before address generation process starts. Those who starts address generation process also sets a time for a current address generation session. This time will be used by multiple nodes.

### DIFF
--- a/btc-address-generation/src/main/kotlin/generation/btc/init/BtcAddressGenerationInitialization.kt
+++ b/btc-address-generation/src/main/kotlin/generation/btc/init/BtcAddressGenerationInitialization.kt
@@ -95,9 +95,8 @@ class BtcAddressGenerationInitialization(
             btcAddressGenerationConfig.registrationAccount.accountId
         ).flatMap { details ->
             // Getting time
-            val time = details[ADDRESS_GENERATION_TIME_KEY]!!.toLong()
-            // Removing generation time. Only public keys remain
-            details.remove(ADDRESS_GENERATION_TIME_KEY)
+            val time = details.remove(ADDRESS_GENERATION_TIME_KEY)!!.toLong()
+            // Getting keys
             val notaryKeys = details.values
             btcPublicKeyProvider.checkAndCreateMultiSigAddress(notaryKeys, addressType, time)
         }
--- a/btc-address-generation/src/main/kotlin/generation/btc/init/BtcAddressGenerationInitialization.kt
+++ b/btc-address-generation/src/main/kotlin/generation/btc/init/BtcAddressGenerationInitialization.kt
@@ -14,9 +14,10 @@ import mu.KLogging
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.stereotype.Component
-import provider.btc.generation.BtcPublicKeyProvider
 import provider.btc.address.BtcAddressType
 import provider.btc.address.getAddressTypeByAccountId
+import provider.btc.generation.ADDRESS_GENERATION_TIME_KEY
+import provider.btc.generation.BtcPublicKeyProvider
 import sidechain.iroha.IrohaChainListener
 import sidechain.iroha.consumer.IrohaNetwork
 import sidechain.iroha.util.getAccountDetails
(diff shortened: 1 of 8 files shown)


btc-address-generation/src/main/kotlin/generation/btc/init/BtcAddressGenerationInitialization.kt
```diff
@@ -91,8 +94,12 @@ class BtcAddressGenerationInitialization(
             sessionAccount,
             btcAddressGenerationConfig.registrationAccount.accountId
         ).flatMap { details ->
+            // Getting time
+            val time = details[ADDRESS_GENERATION_TIME_KEY]!!.toLong()
+            // Removing generation time. Only public keys remain
+            details.remove(ADDRESS_GENERATION_TIME_KEY)
```
Mingela: We can rewrite it without line 98 since `remove`: `@return the previous value associated with the key, or `null` if the key was not present in the map.`

btc/src/main/kotlin/provider/btc/generation/BtcPublicKeyProvider.kt
```diff
@@ -72,48 +73,56 @@ class BtcPublicKeyProvider(
...
                 throw IllegalStateException("No peers to create btc multisignature address")
-            } else if (notaryKeys.size == peers && hasMyKey(notaryKeys)) {
-                val threshold = getSignThreshold(peers)
-                val msAddress = createMsAddress(notaryKeys, threshold)
-                if (!walletFile.wallet.addWatchedAddress(msAddress)) {
-                    throw IllegalStateException("BTC address $msAddress was not added to wallet")
-                }
-                logger.info { "Address $msAddress was added to wallet. Current wallet state:\n${walletFile.wallet}" }
+            } else if (notaryKeys.size != peers) {
+                logger.info { "Not enough keys are collected to generate a multisig address(${notaryKeys.size} out of ${peers})" }
```
Mingela: `${peers}` braces are redundant

btc/src/main/kotlin/provider/btc/generation/BtcPublicKeyProvider.kt
```diff
@@ -72,48 +73,56 @@ class BtcPublicKeyProvider(
...
+                return@of
```
Mingela: Don't we need to throw something instead of just `return@of` ?

btc/src/main/kotlin/provider/btc/generation/BtcPublicKeyProvider.kt
```diff
@@ -72,48 +73,56 @@ class BtcPublicKeyProvider(
...
-                }, { ex -> throw ex })
             }
+            ModelUtil.setAccountDetail(
+                multiSigConsumer,
+                storageAccount,
+                msAddress.toBase58(),
+                addressInfo.toJson(),
+                BigInteger.valueOf(generationTime)
+            ).fold({
+                //TODO this save will probably corrupt the wallet file
```
Mingela: Is it bad? What should we do with that in the future?

grimadas: Can you make a task for that?

btc/src/main/kotlin/provider/btc/generation/BtcPublicKeyProvider.kt
dolgopolovwork: No. It simply means that we in the middle of the process.

btc/src/main/kotlin/provider/btc/generation/BtcPublicKeyProvider.kt
dolgopolovwork: Oh, yeah. You are right. It will be fixed.

btc/src/main/kotlin/provider/btc/generation/BtcPublicKeyProvider.kt
dolgopolovwork: It's very bad. But it also outdated)

PR #328, opened 2018-12-27
add automatic relay deployment logic

Alexey-N-Chernyshov:
Signed-off-by: Alexey Chernyshov <alexey.n.chernyshov@gmail.com>

### Description of the Change
Automatic relay deployment.

### DIFF
--- a/eth-registration/src/main/kotlin/registration/eth/relay/RelayRegistration.kt
+++ b/eth-registration/src/main/kotlin/registration/eth/relay/RelayRegistration.kt
@@ -1,20 +1,31 @@
 package registration.eth.relay
 
 import com.github.kittinunf.result.Result
+import com.github.kittinunf.result.failure
+import com.github.kittinunf.result.map
 import config.EthereumPasswords
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import model.IrohaCredential
 import mu.KLogging
+import provider.eth.EthFreeRelayProvider
 import sidechain.eth.util.DeployHelper
 import sidechain.iroha.consumer.IrohaConsumerImpl
 import sidechain.iroha.consumer.IrohaNetwork
 import sidechain.iroha.util.ModelUtil
 import java.io.File
+import java.util.*
+import kotlin.concurrent.schedule
+import kotlin.concurrent.timer
 
 /**
  * Class is responsible for relay addresses registration.
  * Deploys relay smart contracts in Ethereum network and records it in Iroha.
  */
 class RelayRegistration(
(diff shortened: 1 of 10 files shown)


eth-registration/src/main/kotlin/registration/eth/relay/RelayRegistration.kt
```diff
@@ -69,6 +69,29 @@ class RelayRegistration(
...
+                logger.info { "Relay replenishment triggered" }
+
+                freeRelayProvider.getRelays().map { relays ->
+                    logger.info { "Free relays: ${relays.size}" }
+                    val toDeploy = relayRegistrationConfig.number - relays.size
+                    if (toDeploy > 0)
+                        deploy(toDeploy, relayRegistrationConfig.ethMasterWallet)
+                }.failure { throw it }
+
+                runBlocking { delay(relayRegistrationConfig.replenishmentPeriod * 1000) }
```
dolgopolovwork: Why can't we use `Thread.sleep(relayRegistrationConfig.replenishmentPeriod * 1000) `instead?

eth-registration/src/main/kotlin/registration/eth/relay/RelayRegistration.kt
```diff
@@ -69,6 +69,29 @@ class RelayRegistration(
...
 
+    /**
+     * Run a job that every replenishmentPeriod checks that number from config free relays are present. In case of
+     * lack of free relays deploys lacking amount.
+     */
+    fun runRelayReplenishment(): Result<Unit, Exception> {
+        logger.info { "Run relay replenishment" }
+
+        return Result.of {
+            while (true) {
```
dolgopolovwork: I think we can make it easier. Please, take a look at [ScheduledExecutorService](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html)

Mingela: I think it is not used for purpose since it is about async execution

dolgopolovwork: Do you mean `IrohaNetworkImpl`? I thought it's a thread-safe class

eth-registration/src/main/kotlin/registration/eth/relay/RelayRegistration.kt
Alexey-N-Chernyshov: https://medium.com/@mohakpuri1712/kotlin-coroutines-thread-sleep-vs-delay-63171fe8a24

Alexey-N-Chernyshov: There is explained a difference between Thread.sleep() and delay(). delay() is non-blocking suspend function. However, there is no need to use it.

PR #330, opened 2019-01-16
Transition to IPJ, Iroha rc2 version update

Mingela:
### Description of the Change
Now we don't need to support bindings. Just use and maintain IPJ library.
Iroha version was also updated to the recent release.


deploy/iroha/genesis.block
```diff
@@ -264,11 +272,17 @@
                   "roleName": "registration_service"
                 }
               },
+              {
+                "appendRole": {
+                  "accountId": "btc_registration_service@notary",
+                  "roleName": "client"
```
dolgopolovwork: Looks suspicious. Why we use `client`?  Why not `registration_service`?

deploy/iroha/genesis.block
```diff
@@ -284,11 +298,17 @@
                   "roleName": "registration_service"
                 }
               },
+              {
+                "appendRole": {
+                  "accountId": "mst_btc_registration_service@notary",
+                  "roleName": "client"
```
dolgopolovwork: Why `client`?

deploy/iroha/genesis.block
```diff
@@ -342,11 +362,23 @@
                   "roleName": "tester"
                 }
               },
+              {
+                "appendRole": {
```
dolgopolovwork: Why we need this block? We have all desired permissions in `tester`

Mingela: These question can be answered by a single phrase: Iroha permissions and domains update.
Now a one need to have all the permissions and roles to issue a new account in domain with such role etc

PR #333, opened 2019-01-20
Feature/healthcheck

x3medima17:
### Description of the Change
Adds healtheck endpoint for withdrawal, registration, and refund(notary) service.

### DIFF
--- a/eth-withdrawal/src/main/kotlin/withdrawalservice/WithdrawalServiceEndpoint.kt
+++ b/eth-withdrawal/src/main/kotlin/withdrawalservice/WithdrawalServiceEndpoint.kt
@@ -1,0 +1,53 @@
+package withdrawalservice
+
+import io.ktor.application.call
+import io.ktor.application.install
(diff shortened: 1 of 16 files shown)


eth-withdrawal/src/main/kotlin/withdrawalservice/WithdrawalServiceEndpoint.kt
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Withdrawal HTTP service
+ */
+class WithdrawalServiceEndpoint(
+        port: Int
+) {
+
+    init {
+        RegistrationServiceEndpoint.logger.info { "Start withdrawal server on port $port" }
```
dolgopolovwork: `RegistrationServiceEndpoint ` can be omitted

Mingela:
How about btc services as well?

dolgopolovwork:
>How about btc services as well?
We already have health checks in btc.

PR #344, opened 2019-01-25
Bitcoin-cli in gradle

dolgopolovwork:
### Description of the Change
Now we can send Bitcoins and generate blocks with no bitcoin-cli or bitcoind being involved. 
Usage:
1) Generate blocks `./gradlew btcGenerateBlocks -Pblocks=<blocks>` 
2) Send to address `./gradlew btcSendToAddress -Paddress=<address> -PamountBtc=<amountBtc>` (no need to generate 6 blocks after that)

### DIFF
--- a/btc/src/main/kotlin/cli/GenerateBlocks.kt
+++ b/btc/src/main/kotlin/cli/GenerateBlocks.kt
@@ -1,0 +1,45 @@
+@file:JvmName("BtcGenerateBlocksMain")
+
+package cli
+
+import com.github.jleskovar.btcrpc.BitcoinRpcClientFactory
+import com.github.kittinunf.result.Result
+import config.loadConfigs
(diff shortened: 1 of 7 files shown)


btc/src/main/kotlin/cli/GenerateBlocks.kt
```diff
@@ -0,0 +1,45 @@
...
+    port = btcNodeRpcConfig.port,
+    secure = false
+)
+
+/**
+ * Generates blocks in Bitcoin blockchain
+ * @param args - array full of arguments. args[0] - number of blocks to generate
+ */
+fun main(args: Array<String>) {
+    val blocksToGenerate = args[0].toInt()
```
Mingela: I suggest to add some default value (6) if an argument value is not provided or it's less than 1

PR #346, opened 2019-01-28
Fixing IPJ listener bug

dolgopolovwork:
### Description of the Change
Iroha chain listener handles blocks in grpc thread pool no matter what pool was set by developers. In some services, it's crucial to handle blocks in a single thread. This bug was fixed on IPJ side. We must fix it on our side.


notary-commons/src/main/kotlin/sidechain/iroha/IrohaChainListener.kt
```diff
@@ -13,11 +13,16 @@ import sidechain.iroha.util.ModelUtil
  * Dummy implementation of [ChainListener] with effective dependencies
  */
 class IrohaChainListener(
-    irohaHost: String,
-    irohaPort: Int,
+    private val irohaAPI: IrohaAPI,
```
Mingela: So you should fix all usages as well

PR #350, opened 2019-01-30
Email notifications

dolgopolovwork:
### Description of the Change
Notifications - is a standalone service dedicated to email notifications. It listens to Iroha blocks and notifies clients if deposit or withdrawal event occurred. We assume that client email is set by himself and stored in account details. 
We also need an SMTP server in order to make this service functioning. SMTP configurations are hidden from VCS since it holds SMTP credentials. But you can use [mailtrap.io](http://mailtrap.io) services for testing purposes(or just ask me, because I've already got mailtrap credentials).

### DIFF
--- a/notary-commons/src/main/kotlin/provider/D3ClientProvider.kt
+++ b/notary-commons/src/main/kotlin/provider/D3ClientProvider.kt
@@ -1,0 +1,27 @@
+package provider
+
+import com.github.kittinunf.result.Result
+import com.github.kittinunf.result.map
+import jp.co.soramitsu.iroha.java.QueryAPI
+import model.D3Client
+import sidechain.iroha.util.getAccountDetails
+
+/**
+ * Class that is used to get information about D3 clients
+ */
+class D3ClientProvider(private val notaryQueryAPI: QueryAPI) {
+
+    /**
+     * Returns D3 client by its id
+     * @param accountId - account id of client
+     * @return D3 client
+     */
+    fun getClient(accountId: String): Result<D3Client, Exception> {
+        // Assuming that client sets details himself
+        return getAccountDetails(notaryQueryAPI, accountId, accountId)
+            .map { accountDetails ->
+                D3Client.create(accountId, accountDetails)
+            }
+    }
+}
+
--- a/notifications-integration-test/src/integration-test/kotlin/notifications/NotificationsIntegrationTest.kt
+++ b/notifications-integration-test/src/integration-test/kotlin/notifications/NotificationsIntegrationTest.kt
@@ -1,0 +1,235 @@
+package notifications
+
+import com.github.kittinunf.result.Result
+import com.github.kittinunf.result.failure
+import com.github.kittinunf.result.map
+import com.nhaarman.mockito_kotlin.*
+import integration.helper.IrohaIntegrationHelperUtil
+import model.D3_CLIENT_EMAIL_KEY
+import model.D3_CLIENT_ENABLE_NOTIFICATIONS
+import notifications.environment.NotificationsIntegrationTestEnvironment
+import notifications.service.D3_DEPOSIT_EMAIL_SUBJECT
+import notifications.service.D3_WITHDRAWAL_EMAIL_SUBJECT
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.api.fail
+import org.mockito.Matchers.anyString
+import sidechain.iroha.CLIENT_DOMAIN
+import sidechain.iroha.util.ModelUtil
+import java.math.BigDecimal
+
+const val BTC_ASSET = "btc#bitcoin"
+private const val TRANSFER_WAIT_TIME = 3_000L
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class NotificationsIntegrationTest {
+
+    private val integrationHelper = IrohaIntegrationHelperUtil()
+
+    private val environment = NotificationsIntegrationTestEnvironment(integrationHelper)
+
+    init {
+        // This amount is big enough for testing
+        val amount = BigDecimal(100)
+
+        // Creating 2 clients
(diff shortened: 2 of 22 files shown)


notifications/src/main/kotlin/notifications/service/EmailNotificationService.kt
```diff
@@ -0,0 +1,64 @@
...
+@Component
+class EmailNotificationService(
+    @Autowired private val smtpService: SMTPService,
+    @Autowired private val d3ClientProvider: D3ClientProvider
+) : NotificationService {
+
+    override fun notifyDeposit(transferNotifyEvent: TransferNotifyEvent): Result<Unit, Exception> {
+        logger.info { "Notify deposit $transferNotifyEvent" }
+        val message =
+            "Hello ${transferNotifyEvent.accountId}. We are happy to notify that ${transferNotifyEvent.amount} ${transferNotifyEvent.assetName} has been deposited to your account."
```
Mingela: Hello **dear**
Are we really happy? 🤣 
I think it is better to ask @LiraLemur (https://github.com/LiraLemur) to help with that

notifications/src/main/kotlin/notifications/smtp/SMTPService.kt
```diff
@@ -0,0 +1,64 @@
...
+            })
+        }.map { session ->
+            val mimeMessage = MimeMessage(session)
+            mimeMessage.setFrom(InternetAddress(from))
+            mimeMessage.setRecipients(
+                Message.RecipientType.TO, InternetAddress.parse(to)
+            )
+            mimeMessage.subject = subject
+            val mimeBodyPart = MimeBodyPart()
+            mimeBodyPart.setContent(message, "text/html")
```
Mingela: I know this is the initial version but for future I suggest to use email template, for example, [FreeMarker](https://freemarker.apache.org) (flexible CSS styles support)

notifications/src/main/kotlin/notifications/service/EmailNotificationService.kt
LiraLemur: Is is the final state? I think something luke "Dear _. N tokens were deposited to your account." would be sufficient :) Though I am not sure about the context

dolgopolovwork: Oh, thank you @LiraLemur . Can you please also take a look at withdrawal message?

notary-commons/src/main/kotlin/provider/D3ClientProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+class D3ClientProvider(private val notaryQueryAPI: QueryAPI) {
+
+    /**
+     * Returns D3 client by its id
+     * @param accountId - account id of client
+     * @return D3 client
+     */
+    fun getClient(accountId: String): Result<D3Client, Exception> {
+        // Assuming that client sets details himself
```
Mingela: hmm it's needed to be decided
probably `registration_service` could also take care of it

notifications-integration-test/src/integration-test/kotlin/notifications/NotificationsIntegrationTest.kt
```diff
@@ -0,0 +1,235 @@
...
+            D3_CLIENT_EMAIL_KEY,
+            "user@d3.com"
+        ).failure { ex -> throw ex }
+
+        // Setting dest client email
+        ModelUtil.setAccountDetail(
+            environment.destClientConsumer,
+            environment.destClientId,
+            D3_CLIENT_EMAIL_KEY,
+            "user@d3.com"
```
Mingela: just to notice you are using the same email for both clients

notifications/src/main/kotlin/notifications/init/NotificationInitialization.kt
```diff
@@ -0,0 +1,100 @@
...
+                }, { ex ->
+                    logger.error("Error on Iroha subscribe", ex)
+                    onIrohaChainFailure()
+                })
+        }
+    }
+
+    // Checks if deposit event
+    private fun isDeposit(transferAsset: Commands.TransferAsset): Boolean {
+        return transferAsset.srcAccountId.endsWith("@$NOTARY_DOMAIN")
```
Mingela: Do we need to include `@`? Are you assuming we are going to have a domain like `@blablablanotary`? In that case it is better to assign `@notary` to `NOTARY_DOMAIN` unless it is used somewhere else separately

notifications/src/main/kotlin/notifications/service/EmailNotificationService.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.github.kittinunf.result.Result
+import com.github.kittinunf.result.flatMap
+import mu.KLogging
+import notifications.smtp.SMTPService
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+import provider.D3ClientProvider
+
+private const val NOTIFICATION_EMAIL = "noreply@d3.com"
+const val D3_WITHDRAWAL_EMAIL_SUBJECT = "D3 withdrawal"
```
Mingela: I think it is common practice just to use the same subject name for all notifications. Smth like `D3 notification` should be fine

Mingela:
Also could you provide some sample `smtp.properties` ?

notifications/src/main/kotlin/notifications/init/NotificationInitialization.kt
dolgopolovwork: Yes, I think that `@blablanotarty` may happen. `NOTARY_DOMAIN` represents the domain. `notary` is a valid domain, while `@notary` is not.

notary-commons/src/main/kotlin/provider/D3ClientProvider.kt
dolgopolovwork: Unfortunately, we use different `registration_service` accounts for Bitcoin and Ethereum. I suppose that crypto related details will be set by `registration_service` while common information (such as email) will be set by user himself.

PR #357, opened 2019-02-05
Release/v0.2.1 alpha - merge to develop

Alexey-N-Chernyshov:
### Description of the Change
Backward merge - to be able merge without conflicts in future


Alexey-N-Chernyshov:
Seems formatting was done by other version of solc. I think we should keep it.

PR #377, opened 2019-02-18
Splitting Iroha chain listener

dolgopolovwork:
### Description of change
Now Iroha chain listener is split into two: plain Iroha chain listener that listens to block directly from Iroha and reliable chain listener that listens to Rabbit MQ

### DIFF
--- a/notary-commons/src/main/kotlin/sidechain/iroha/ReliableIrohaChainListener.kt
+++ b/notary-commons/src/main/kotlin/sidechain/iroha/ReliableIrohaChainListener.kt
@@ -10,7 +10,6 @@ import io.reactivex.subjects.PublishSubject
 import jp.co.soramitsu.iroha.java.IrohaAPI
 import mu.KLogging
 import sidechain.ChainListener
-import kotlin.test.assertNotNull
 
 /**
  * Rabbit MQ based implementation of [ChainListener]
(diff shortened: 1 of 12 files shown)


notary-commons/src/main/kotlin/sidechain/iroha/ReliableIrohaChainListener.kt
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    /**
+     * @return a block as soon as it is committed to iroha
+     */
+    override suspend fun getBlock(autoAck: Boolean): Pair<iroha.protocol.BlockOuterClass.Block, () -> Unit> {
+        var resp: GetResponse?
+        do {
+            resp = channel.basicGet(irohaQueue, autoAck)
```
x3medima17: Please add some sleep for ~10ms, i forgot

dolgopolovwork: Before or after this line?

x3medima17: Doesn't matter :D
I would add it after

dolgopolovwork: Ok. I'll do this. But what is the purpose of that?

x3medima17: Here we have busy waiting, which can load CPU

PR #381, opened 2019-02-19
Feature/upgradable contracts

Alexey-N-Chernyshov:
### Description of the Change
Upgradable smart contracts.
Upgradable strategy is implemented with proxy mechanism (https://soramitsu.atlassian.net/wiki/spaces/ESC/pages/683737089/Upgradable+Strategy).

##### Changes in contracts
Since proxy separates implementation and storage, we need explicitly call initialise in upgradable contracts.
The following contracts now are deployed with proxy:
- Matster
- RelayRegistry
- Relay


deploy/ethereum/contract/Relay.sol
```diff
@@ -74,7 +74,7 @@ contract Relay {
     function withdraw(
         address tokenAddress,
         uint256 amount,
-        address to,
+        address payable to,
```
Hirama: We still can withdraw tokens? Looks like we bind to send only ethers

PR #386, opened 2019-02-22
Feature/add xor erc20

Alexey-N-Chernyshov:
### Description of the Change
Add Sora token with initial supply.
Please, pay attention that tokens in token list is now stored with domain (xor#sora) since we need sora domain for minting restrictions. Previously we assumed all tokens to have ethereum domain (omg#ethereum).


notary-eth-integration-test/src/integration-test/kotlin/integration/eth/ERC20TokenRegistrationTest.kt
```diff
@@ -57,14 +61,23 @@ class ERC20TokenRegistrationTest {
             executeTokenRegistration(tokenRegistrationConfig)
             ethTokensProvider.getTokens()
                 .map { tokensFromProvider ->
-                    val expected = tokensFromProvider
-                        .map { (ethAddress, name) ->
-                            Pair(ethAddress, EthTokenInfo(name, ethTokensProvider.getTokenPrecision(name).get()))
+                    val actual = tokensFromProvider
+                        .map { (ethAddress, tokenId) ->
+                            println(tokenId)
```
dolgopolovwork: I think it must be properly logger using KLogging

notary-eth-integration-test/src/integration-test/kotlin/integration/eth/ERC20TokenRegistrationTest.kt
```diff
@@ -57,14 +61,21 @@ class ERC20TokenRegistrationTest {
             executeTokenRegistration(tokenRegistrationConfig)
             ethTokensProvider.getTokens()
                 .map { tokensFromProvider ->
-                    val expected = tokensFromProvider
-                        .map { (ethAddress, name) ->
-                            Pair(ethAddress, EthTokenInfo(name, ethTokensProvider.getTokenPrecision(name).get()))
+                    val actual = tokensFromProvider
+                        .map { (ethAddress, tokenId) ->
+                            val name = tokenId.split("#").first()
```
Mingela: redundant splitting, we can declare a variable before it, btw can we do smth like `(first, last) = split` in Kotlin?

Alexey-N-Chernyshov: Split returns list and list cannot be used in destructuring declarations.

PR #392, opened 2019-02-26
Fixed wallet storage

dolgopolovwork:
### Description of the Change
Wallet object can do many things:
1) Send money
2) Receive money
3) Store private keys
4) Keep information about UTXO
5) Watch for particular address activity
6) Dozens of listeners

And we also need to save this file periodically to keep wallet data persistent. Unfortunately, due to D3 microservice nature, it's not so easy as it sounds, because this wallet is shared among two JVM processes(withdrawal and address generation). Concurrent save will corrupt or/and break wallet consistency by overriding previously written data. 
The easiest (and also very stable) solution is having two separate wallet files: one for keeping data about transfers/UTXO(`transfers.d3.wallet`) and one for priv/pub keys storage(`keys.d3.wallet`).


.gitignore
```diff
@@ -146,12 +146,10 @@ deploy/ethereum/network
 deploy/ethereum/keys/test
 
 # Bitcoin wallets, blockchain and etc.
-deploy/bitcoin/testnet/d3.wallet
 deploy/bitcoin/testnet/blocks/
 deploy/bitcoin/regtest/blocks/
-deploy/bitcoin/regtest/d3.wallet
-deploy/bitcoin/testnet/d3.wallet
-deploy/bitcoin/regtest/d3-test.wallet
+deploy/bitcoin/regtest/keys.d3-test.wallet
```
Mingela: what about testnet?

btc-deposit/src/main/kotlin/com/d3/btc/deposit/listener/BitcoinTransactionListener.kt
```diff
@@ -17,20 +17,27 @@ private const val BTC_ASSET_NAME = "btc"
...
     private val confidenceLevel: Int,
-    // Source of Bitcoin deposit events
     private val emitter: ObservableEmitter<SideChainEvent.PrimaryBlockChainEvent>,
-    // Executor that will be used to execute confidence listener logic
-    private val confidenceListenerExecutor: ExecutorService
+    private val confidenceListenerExecutor: ExecutorService,
+    private val onUnspent: () -> Unit
 ) {
     fun onTransaction(tx: Transaction, blockTime: Date) {
+        //TODO save change address
```
Mingela: what is that about?

btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/config/BtcWithdrawalConfig.kt
```diff
@@ -31,4 +31,8 @@ interface BtcWithdrawalConfig {
     val notaryListSetterAccount: String
     // RabbitMQ queue that is used for listening to Iroha blocks
     val irohaBlockQueue: String
+    // Path to wallet that stores transfers(UTXO)
+    val btcTransfersWalletPath: String
```
Mingela: Maybe these should be added to `BitcoinConfig` ?

notary-commons/src/main/kotlin/config/Configs.kt
```diff
@@ -71,8 +71,6 @@ interface EthereumConfig {
  * Bitcoin configurations
  */
 interface BitcoinConfig {
-    //Path of wallet file
-    val walletPath: String
```
Mingela: why not to introduce 2 wallets here?

dolgopolovwork: Some services don't need two wallets. Like `btc-address-generation`

btc-deposit/src/main/kotlin/com/d3/btc/deposit/listener/BitcoinTransactionListener.kt
dolgopolovwork: We need to save the wallet when change(noun) address is deposited.

.gitignore
dolgopolovwork: You are right. My bad.

notary-btc-integration-test/src/integration-test/kotlin/integration/btc/BtcMultiWithdrawalIntegrationTest.kt
```diff
@@ -56,10 +56,11 @@ class BtcMultiWithdrawalIntegrationTest {
             }
         val someEnvironment = environments.first()
         integrationHelper.generateBtcInitialBlocks()
-        integrationHelper.genChangeBtcAddress(someEnvironment.btcWithdrawalConfig.bitcoin.walletPath)
+        //TODO don't forget to add chain address to transfers wallet
```
Alexey-N-Chernyshov: Is it relevant?

dolgopolovwork: This is very relevant. Will be fixed in next pr.

PR #411, opened 2019-03-06
Added clients accounts to the notary account

Mingela:
### Description of the Change
Added saving of created user accounts to 'notary' account in a format:
`usernamedomain -> domain`

### DIFF
--- a/configs/registration_local.properties
+++ b/configs/registration_local.properties
@@ -5,7 +5,8 @@ registration.port=8085
 registration.registrationCredential.accountId=registration_service@notary
 registration.registrationCredential.pubkeyPath=deploy/iroha/keys/registration_service@notary.pub
 registration.registrationCredential.privkeyPath=deploy/iroha/keys/registration_service@notary.priv
-# --------- Iroha --------------
+registration.clientStorageAccount=notary@notary
+# --------- Iroha ---------
 # Iroha peer hostname
 registration.iroha.hostname=d3-iroha
 # Iroha peer port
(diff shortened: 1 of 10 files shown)


Mingela:
We should probably eliminate side chain-specific writes after transition to the Iroha account creation as the only way of actual registering

notary-registration/src/main/kotlin/registration/NotaryRegistrationStrategy.kt
```diff
@@ -13,7 +13,8 @@ import sidechain.iroha.util.ModelUtil
  */
 @Component
 class NotaryRegistrationStrategy(
-    @Autowired private val irohaConsumer: IrohaConsumer
+    @Autowired private val irohaConsumer: IrohaConsumer,
+    @Autowired private val notaryAccount: String
```
dolgopolovwork: I'd rather add `@Qualifier`. `String` is a very popular class. Somebody may create another `String` bean by mistake.

notary-registration/src/main/kotlin/registration/NotaryRegistrationAppConfiguration.kt
```diff
@@ -41,4 +42,7 @@ class NotaryRegistrationAppConfiguration {
     @Bean
     fun registrationConfig() = registrationConfig
 
+    @Bean
+    @Qualifier("notaryAccount")
```
dolgopolovwork: No need to do it here. Bean is called after `notaryAccount()` function name

configs/registration_local.properties
```diff
@@ -5,6 +5,7 @@ registration.port=8085
 registration.registrationCredential.accountId=registration_service@notary
 registration.registrationCredential.pubkeyPath=deploy/iroha/keys/registration_service@notary.pub
 registration.registrationCredential.privkeyPath=deploy/iroha/keys/registration_service@notary.priv
+registration.notaryAccount=notary@notary
```
Alexey-N-Chernyshov: I would name it as "registration .client_storage_account". Notary is not meaningful.

Mingela:
Will be combined with overall brvs integration PR

PR #412, opened 2019-03-06
Feature/registration docs

Alexey-N-Chernyshov:
### Description of the Change
Update registration docs
Return clientId on registration request


doc/api-client.md
```diff
@@ -18,23 +18,35 @@ That's why D3 services provide registration of new client via HTTP POST request.
 * **Parameters:**
 
     * ***name*** - client name
-    * ***whitelist*** - comma separated list of addresses where withdrawal is possible
```
Mingela: it should be removed from the method signature as well

PR #416, opened 2019-03-07
add docker-wait-for d3-rmq

Alexey-N-Chernyshov:
### Description of the Change
Add wait for d3-rmq and d3-iroha on chain-adapter startup


Alexey-N-Chernyshov:
Yep, we can remove grpc_healthcheck.

PR #417, opened 2019-03-11
Renaming notary to deposit

dolgopolovwork:
### Description of the Change
Renaming Ethereum notary to deposit


eth/src/test/kotlin/com/d3/eth/notary/endpoint/RefundServerEndpointTest.kt
```diff
@@ -9,6 +9,7 @@ import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
 import com.d3.commons.notary.endpoint.ServerInitializationBundle
+import com.d3.eth.deposit.endpoint.*
```
Mingela: seems strange

PR #420, opened 2019-03-12
Feature/registration

Alexey-N-Chernyshov:
https://soramitsu.atlassian.net/browse/D3-765

### Description of the Change
Rework registration services.
Check if client was already registered in sidechain, remove Iroha account registration in sidechain registration.

### Usage Examples or Tests *[optional]*
Update frontend before merge.


eth-registration/src/main/kotlin/com/d3/eth/registration/EthRegistrationStrategyImpl.kt
```diff
@@ -47,33 +55,51 @@ class EthRegistrationStrategyImpl(
...
         credentials,
-        BigInteger.valueOf(ethRegistrationConfig.ethereum.gasPrice),
-        BigInteger.valueOf(ethRegistrationConfig.ethereum.gasLimit)
+        StaticGasProvider(
+            BigInteger.valueOf(ethRegistrationConfig.ethereum.gasPrice),
+            BigInteger.valueOf(ethRegistrationConfig.ethereum.gasLimit)
+        )
     )
 
     private val irohaEthAccountCreator =
```
Mingela: This one became unused

eth-registration/src/main/kotlin/com/d3/eth/registration/EthRegistrationStrategyImpl.kt
```diff
@@ -47,33 +55,51 @@ class EthRegistrationStrategyImpl(
...
+                        // check that client hasn't been registered yet
+                        if (!assignedRelays.isEmpty())
+                            throw IllegalArgumentException("Client $accountName@$domainId has already been registered with relay: $assignedRelays")
+
+                        // register to Ethereum RelayRegistry
+                        logger.info { "Add new relay to relay registry relayRegistry=$relayRegistry, freeWallet=$freeEthWallet, whitelist=$whitelist, creator=${credentials.address}." }
+                        relayRegistry.addNewRelayAddress(freeEthWallet, whitelist).send()
+
+                        // register to Iroha
+                        irohaAccountCreator.create(
```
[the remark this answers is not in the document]
Alexey-N-Chernyshov: User should have account in Iroha in order to register in Ethereum sidechain.
Updated comment.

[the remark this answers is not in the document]
Mingela: Sure, but I'd also recommend to rename `create` and `createAccountCreationBatch` methods to avoid confusion. For example, we could use some sort of `set registered in side chain information in Iroha` terminology.

eth/src/main/kotlin/com/d3/eth/provider/EthRelayProviderIrohaImpl.kt
```diff
@@ -40,9 +41,13 @@ class EthRelayProviderIrohaImpl(
     }
 
     /** Get relay belonging to [irohaAccountId] */
-    override fun getRelay(irohaAccountId: String): Result<String, Exception> {
+    override fun getRelayByAccountId(irohaAccountId: String): Result<Optional<String>, Exception> {
         return getRelays().map { relays ->
-            relays.filter { it.value == irohaAccountId }.keys.first()
+            val filtered = relays.filter { it.value == irohaAccountId }.keys
+            if (filtered.isEmpty())
+                Optional.empty()
```
dolgopolovwork: Why we use `Optional` here? `String?` reduces the need for that.

notary-btc-integration-test/build.gradle
```diff
@@ -46,6 +46,7 @@ dependencies {
     compile project(":btc-address-generation")
     compile project(":btc-deposit")
     compile project(":btc-withdrawal")
+    compile project(":notary-registration")
```
dolgopolovwork: Why we need it here?

eth/src/main/kotlin/com/d3/eth/provider/EthRelayProviderIrohaImpl.kt
Alexey-N-Chernyshov: Good point, but unfortunately:
1. Kotlin Result cannot be used as return from function right now (https://stackoverflow.com/questions/52631827/why-cant-kotlin-result-be-used-as-a-return-type)
2. So, we use https://github.com/kittinunf/Result
3. The last result has type Result<Any, Exception> but not Result<Any?, Exception>, so nullable type cannot be used.
So, only option I see is to use Java Optional.

PR #431, opened 2019-03-15
Fixed batch check

dolgopolovwork:
### Description of the Change
There was an additional `CreateAccount` command in `IrohaAccountRegistrator.register()` that could fail legally. But it's not the case anymore, so `isAccountCreationBatchSuccessful` must be fixed. Now it just checks if ALL commands in a batch are successful.

### DIFF
--- a/notary-commons/src/main/kotlin/com/d3/commons/registration/IrohaAccountRegistrator.kt
+++ b/notary-commons/src/main/kotlin/com/d3/commons/registration/IrohaAccountRegistrator.kt
@@ -116,18 +116,15 @@ open class IrohaAccountRegistrator(
     }
 
     /**
-     * Checks if account creation batch successful
+     * Checks if account creation batch is successful
      * @param transactionsResult - map of processed transactions and the indicator if it was successful
-     * @return 'true' if all [transactionsResult] (except for first transaction, we can ignore it) are successful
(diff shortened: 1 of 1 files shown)


notary-commons/src/main/kotlin/com/d3/commons/registration/IrohaAccountRegistrator.kt
```diff
@@ -116,18 +116,15 @@ open class IrohaAccountRegistrator(
...
     protected open fun isAccountCreationBatchSuccessful(
         transactionsResult: Map<String, Boolean>
     ): Boolean {
-        val count = transactionsResult.count { it.value }
-        return if (!transactionsResult.values.first()) {
-            count == transactionsResult.size - 1
-        } else
-            count == transactionsResult.size
+        val successfulCommands = transactionsResult.count { it.value }
+        return successfulCommands == transactionsResult.size
```
Mingela: how about `return transactionsResult.values.all { it }`

Mingela: 1 more point: let's move to `ATOMIC BATCH`?

dolgopolovwork: It's a good point

PR #432, opened 2019-03-17
Restart unconfirmed transactions

dolgopolovwork:
### Description of the Change
This is how deposit service works:
1) Listen to Bitcoin blockchain
2) Add depth listener to transactions that we are interested in (deposits to our clients)
3) Wait for 6 blocks
4)  Publish deposit event

If deposit service fails before transaction hits 6 blocks, we will lose money in Iroha after service restart because there will be no listeners for unconfirmed transactions. We have to manually check if we have unconfirmed transactions and reset depth listeners once again.

### DIFF
--- a/btc-deposit/src/main/kotlin/com/d3/btc/deposit/handler/BtcDepositTxHandler.kt
+++ b/btc-deposit/src/main/kotlin/com/d3/btc/deposit/handler/BtcDepositTxHandler.kt
@@ -1,0 +1,70 @@
+package com.d3.btc.deposit.handler
+
+import com.d3.btc.helper.address.outPutToBase58Address
+import com.d3.btc.helper.currency.satToBtc
+import com.d3.btc.model.BtcAddress
+import com.d3.commons.sidechain.SideChainEvent
+import io.reactivex.subjects.PublishSubject
+import mu.KLogging
+import org.bitcoinj.core.Transaction
+import java.math.BigInteger
+import java.util.*
+
+private const val BTC_ASSET_NAME = "btc"
+private const val BTC_ASSET_DOMAIN = "bitcoin"
+private const val TWO_HOURS_MILLIS = 2 * 60 * 60 * 1000L
+
+/**
+ * Handler of Bitcoin deposit transactions
+ * @param registeredAddresses - registered addresses. Used to get Bitcoin address data
+ * @param btcEventsSource - source of Bitcoin deposit events
+ * @param onTxSave - function that is called to save transaction in wallet
+ */
+class BtcDepositTxHandler(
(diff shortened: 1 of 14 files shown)


btc-deposit/src/main/kotlin/com/d3/btc/deposit/handler/BtcDepositTxHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+     * Handles deposit transaction
+     * @param tx - Bitcoin deposit transaction
+     * @param blockTime - time of block where [tx] appeared for the first time. This time is used in MST
+     */
+    fun handleTx(tx: Transaction, blockTime: Date) {
+        tx.outputs.forEach { output ->
+            val txBtcAddress = outPutToBase58Address(output)
+            logger.info { "Tx ${tx.hashAsString} has output address $txBtcAddress" }
+            val btcAddress = registeredAddresses.firstOrNull { btcAddress -> btcAddress.address == txBtcAddress }
+            if (btcAddress != null) {
```
Mingela: what should be done if there is no such addresses? just nothing?

btc-deposit/src/main/kotlin/com/d3/btc/deposit/service/BtcWalletListenerRestartService.kt
```diff
@@ -0,0 +1,114 @@
...
+        registeredAddresses: List<BtcAddress>,
+        onTxSave: () -> Unit
+    ) {
+        // Get tx block hash
+        unconfirmedTx.appearsInHashes?.let { appearsInHashes ->
+            appearsInHashes.keys.firstOrNull()?.let { blockHash ->
+                // Get tx block by hash
+                peerGroup.getBlock(blockHash)?.let { block ->
+                    // Create listener
+                    unconfirmedTx.confidence.addEventListener(
```
Mingela: Can we guarantee there is not more than one such listener attached to the tx?

btc-deposit/src/main/kotlin/com/d3/btc/deposit/handler/BtcDepositTxHandler.kt
dolgopolovwork: Yes. It means that `txBtcAddress` is not our client address.

btc-deposit/src/main/kotlin/com/d3/btc/deposit/service/BtcWalletListenerRestartService.kt
dolgopolovwork: No.

PR #445, opened 2019-03-24
Btc UTXO consensus

dolgopolovwork:
### Description of the Change
Bitcoin withdrawal works this way:
1) Every node creates Bitcoin transaction on Iroha transfer event
2) Every node signs transaction
3) Once a transaction is properly signed, it will be sent to Bitcoin network

Unfortunately, there lies a problem. Every node must pick the same set of UTXO in order to create the same transaction. Due to synchronization issues, different UTXO may be picked. So a mechanism to help nodes to decide what UTXO can be picked was developed. It works very simply:
1) Every node sends some information about its state before creating Bitcoin transaction. For example, last seen Bitcoin block height.
2) Once enough information is collected, nodes may decide what UTXO will be picked

We call it 'withdrawal consensus'. Currently, consensus holds information about peers and last available Bitcoin block height.


btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/config/BtcWithdrawalConfig.kt
```diff
@@ -13,8 +13,11 @@ interface BtcWithdrawalConfig {
     val signatureCollectorCredential: IrohaCredentialConfig
     // Account that is used to deal with registered accounts
     val registrationCredential: IrohaCredentialConfig
+    //TODO this must be removed
```
Alexey-N-Chernyshov: remove?

dolgopolovwork: This part of the config is redundant.  It will be removed soon,

PR #446, opened 2019-03-25
Feature/task deposit from contract

Alexey-N-Chernyshov:
### Description of the Change
Add Ethereum contract to send ether with internal transaction for testing purpose.


eth/src/main/kotlin/com/d3/eth/deploy/TransferEthereumFromContract.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.github.kittinunf.result.map
+import mu.KLogging
+import java.math.BigInteger
+
+private val logger = KLogging().logger
+
+/**
+ * Task that transfer Ethereum from contract as internal transaction for testing purpose.
+ */
+fun main(args: Array<String>) {
```
dolgopolovwork: I'd rather add some information about arguments.

Issue #454, opened 2019-03-26
Missing role definition in Genesis Block produced by Bootstrap service


bakhtin:
Genesis Block (GB) produced by /iroha/create/genesisBlock Bootstrap service endpoint contains a role without its definition (`createRole` command is missing). GB snippet looks the following:
```
...
{
  "createDomain": {
    "defaultRole": "sora_client",
    "domainId": "sora"
  }
}
...
```
`sora_client` role here is a non-existent one. Attempting to start Iroha with that GB leads to the following error:
```
[2019-03-25 13:21:58.962607202][I][Irohad/Storage]: drop blocks from disk
[2019-03-25 13:21:58.962674591][I][Irohad/Storage]: create mutable storage
[2019-03-25 13:21:58.962780289][I][Irohad/Storage]: get(0) file not found
[2019-03-25 13:21:58.964027486][I][Irohad/Storage/MutableStorageImpl]: Applying block: height 1, hash 53272a433287e25339c75fcbb07bd9c34ccc325de3ccabba5d51c290b8f4e25d
[2019-03-25 13:21:58.972850742][E][Irohad/Storage/MutableStorageImpl]: CreateDomain: 4 with extra info 'Query arguments: [domain_id=sora, default_role=sora_client, ]'
[2019-03-25 13:21:58.973205103][I][Irohad/Storage]: block inserted: false
[2019-03-25 13:21:58.974682279][I][Irohad/FlatFileBlockFactory]: Remove /tmp/9e14-6a81-fdb2-f846 temp directory
[2019-03-25 13:21:58.975145689][I][Init]: Genesis block inserted, number of transactions: 1
[2019-03-25 13:21:58.975346539][I][Irohad/Storage]: get(0) file not found
[2019-03-25 13:21:58.975365369][E][Init]: Unable to start the ledger. There are no blocks in the ledger. Please ensure that you are not trying to start the newer version of the ledger over incompatible version of the storage or there is enough disk space. Try to specify --genesis_block and --overwrite_ledger parameters at the same time.
iroha-0-97d3b87b exited with code 1
```

Issue #453, opened 2019-03-26
Bootstrap service creates roles with conflicting names in a Genesis Block


bakhtin:
Genesis Block (GB) produced by `/iroha/create/genesisBlock` Bootstrap service endpoint contains multiple conflicting role names created by `createRole` command. The resulting GB snippet looks the following:
```
...
{
    "createRole": {
        "roleName": "notary_list_holder",
        "permissions": [
            "can_set_detail"
        ]
    }
},
{
    "createRole": {
        "roleName": "notary_list_holder",
        "permissions": [
            "can_get_my_account",
            "can_get_my_acc_ast",
            "can_get_my_acc_ast_txs",
            "can_get_my_acc_txs",
            "can_get_my_txs",
            "can_transfer",
            "can_receive",
            "can_set_quorum",
            "can_add_signatory",
            "can_get_my_signatories",
            "can_remove_signatory",
            "can_grant_can_set_my_quorum",
            "can_grant_can_add_my_signatory",
            "can_grant_can_remove_my_signatory"
        ]
    }
},
{
    "createRole": {
        "roleName": "notary_list_holder",
        "permissions": [
            "can_set_detail",
            "can_get_all_accounts"
        ]
    }
},
{
    "createRole": {
        "roleName": "notary_list_holder",
        "permissions": [
            "can_get_my_acc_ast",
            "can_transfer",
            "can_receive",
            "can_add_domain_asset_qty"
        ]
    }
}
...
```
Attempting to start Iroha leads to the following error:
```
[2019-03-25 13:10:33.992482482][I][Irohad/Storage]: drop blocks from disk
[2019-03-25 13:10:33.992555294][I][Irohad/Storage]: create mutable storage
[2019-03-25 13:10:33.992585002][I][Irohad/Storage]: get(0) file not found
[2019-03-25 13:10:33.993731646][I][Irohad/Storage/MutableStorageImpl]: Applying block: height 1, hash b8380cd9a5d26cade0614970a354cfb8292fcfd21214d6f06964b228d71ca6ce
[2019-03-25 13:10:33.999315456][E][Irohad/Storage/MutableStorageImpl]: CreateRole: 3 with extra info 'Query arguments: [role_id=notary_list_holder, perm_str=000001110100100100000100100100011000111000000, ]'
[2019-03-25 13:10:33.999648316][I][Irohad/Storage]: block inserted: false
[...]

PR #456, opened 2019-03-26
Bootstrap service

yurivin:
- github #453 Bootstrap service creates roles with conflicting names in a Genesis Block

- github #454 Missing role definition in Genesis Block produced by Bootstrap service


dolgopolovwork:
It would be very nice to call PR more meaningful. I've seen several PRs named 'Bootstrap service'. It's quite confusing.

PR #458, opened 2019-03-26
Fixed Bitcoin withdrwal checks

dolgopolovwork:
### Description of the Change
Bitcoin withdrawal service works this way:
1) Create 'withdrawal consensus'
2) Check if withdrawal is valid
3) Create Bitcoin withdrawal transaction 
4) And so on...

Steps 1 and 2 must be swapped because we may create 'withdrawal consensus' for invalid withdrawal operation.

### DIFF
--- a/btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
+++ b/btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
@@ -50,7 +50,6 @@ class NewTransferHandler(
                     "hash:${withdrawalDetails.irohaFriendlyHashCode()})"
         }
 
-        // Check withdrawal
         if (!CurrentFeeRate.isPresent()) {
             logger.warn { "Cannot execute transfer. Fee rate was not set." }
             btcRollbackService.rollback(sourceAccountId, satAmount, withdrawalTime, "Not able to transfer yet")
--- a/btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
+++ b/btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
@@ -1,9 +1,13 @@
 package com.d3.btc.withdrawal.handler
 
+import com.d3.btc.helper.address.isValidBtcAddress
 import com.d3.btc.helper.currency.btcToSat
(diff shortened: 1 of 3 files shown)


btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
```diff
@@ -34,7 +41,59 @@ class NewTransferHandler(
...
+        // Check if withdrawal has valid destination address
+        if (!isValidBtcAddress(destinationAddress)) {
+            logger.warn { "Cannot execute transfer. Destination $destinationAddress is not a valid base58 address." }
+            btcRollbackService.rollback(sourceAccountId, satAmount, withdrawalTime, "Invalid address")
+            return
+        }
+        // Check if withdrawal amount is not too little
+        if (transactionHelper.isDust(satAmount)) {
+            btcRollbackService.rollback(sourceAccountId, satAmount, withdrawalTime, "Too small amount")
+            logger.warn { "Can't spend SAT $satAmount, because it's considered a dust" }
```
Mingela: I think this line should be before rollbacking

btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
```diff
@@ -34,7 +41,59 @@ class NewTransferHandler(
...
+                        "Cannot withdraw to $destinationAddress, " +
+                                "because it's not in ${withdrawalDetails.sourceAccountId} whitelist"
+                    }
+                }
+            }, { ex ->
+                btcRollbackService.rollback(withdrawalDetails, "Cannot get white list")
+                withdrawalStatistics.incFailedTransfers()
+                logger.error("Cannot check ability to withdraw", ex)
+            })
+
```
Mingela: empty line

btc-withdrawal/src/main/kotlin/com/d3/btc/withdrawal/handler/NewTransferHandler.kt
```diff
@@ -16,9 +20,12 @@ import java.math.BigDecimal
  */
 @Component
 class NewTransferHandler(
```
Mingela: So this class only checks if a transfer is the withdrawal trigger, shouldn't it be renamed accordingly?

PR #462, opened 2019-03-27
Fixed quorum

dolgopolovwork:
### Description of the Change
Quorum value must be taken from Iroha account, not from `peerListProvider` because it returns the number of peers. Given that fact, `peerListProvider` is not needed in deposit services.

### DIFF
--- a/notary-commons/src/main/kotlin/com/d3/commons/sidechain/iroha/consumer/IrohaConsumerImpl.kt
+++ b/notary-commons/src/main/kotlin/com/d3/commons/sidechain/iroha/consumer/IrohaConsumerImpl.kt
@@ -53,9 +53,7 @@ class IrohaConsumerImpl(
 
     /**
      * Send transaction to Iroha and check if it is committed
-     * @param utx queryResponse ->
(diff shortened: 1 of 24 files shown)


notary-commons/src/main/kotlin/com/d3/commons/sidechain/iroha/consumer/IrohaConsumerImpl.kt
```diff
@@ -45,13 +43,19 @@ class IrohaConsumerImpl(
...
 
-    val keypair = irohaCredential.keyPair
+    private val keypair = irohaCredential.keyPair
+
+    override fun getConsumerQuorum() = getAccountQuorum(queryAPI, creator)
 
     /**
      * Send transaction to Iroha and check if it is committed
-     * @param utx - unsigned transaction to send
+     * @param utx queryResponse ->
```
Mingela: what is that

dolgopolovwork: oops

Issue #433, opened 2019-03-18
Master contract deployment with Bootstrap service


bakhtin:
Bootstrap service should have a functionality for deploying master contract in Ethereum. It should support deploying in both testnet and mainnet. Endpoint must support passing credentials (in form of a wallet?) to it. The design of the endpoint can be discussed here.

PR #467, opened 2019-03-28
#433 Deploy master & relay smart contracts by Bootstrap service and other fixes

yurivin:


bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/IrohaController.kt
```diff
@@ -130,7 +130,7 @@ class IrohaController(val genesisFactories: List<GenesisInterface>) {
     }
 
     private fun isValidRequest(request: GenesisRequest): Conflictable? {
-        val result = request.peers.stream().filter { it.peerKey == null || it.peerKey.isEmpty() }.collect(toList())
+        val result = request.peers.stream().filter { it.peerKey.isEmpty() }.collect(toList())
```
dolgopolovwork: No need to use `stream()`

bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/dto/Model.kt
```diff
@@ -23,10 +25,48 @@ enum class BtcNetwork(val params: NetworkParameters) {
 
 open class Conflictable(var errorCode: String? = null, var message: String? = null)
 
+
+data class MasterContractsRequest(
+    @NotNull val network: EthereumNetworkProperties? = null,
```
dolgopolovwork: Strange combination of `@NotNull` and `EthereumNetworkProperties? = null`. Is everything ok?

yurivin: Yes, Because @NotNull is for API call, and null is for default constructor, which used by Spring for serialization.

bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/IrohaController.kt
yurivin: fixed

PR #481, opened 2019-04-03
Bootstrap service: Fix failed tests after new accounts added

yurivin:


yurivin:
Looks like there were conflict with other pr which affects tests. This pull request is to fix it.

PR #482, opened 2019-04-04
Updated cfg4k env issue since it got fixed

Mingela:
### Description of the Change
The library maintainer has fixed the issue recently


bakhtin:
I would suggest to wait until the stable version of the library with the fix is released. I already implemented a workaround.

PR #484, opened 2019-04-04
Bootstrap history

dolgopolovwork:
### Description of the Change
Now we are able to store changelog history in Iroha. Every changelog transaction is accompanied by changelog history transaction. This transaction combination forms an atomic batch. All the information is stored in `changelog_history@bootstrap` details (`shemaVersion`->`changelog reducedTxHash`)

### DIFF
--- a/bootstrap/src/integration-test/resources/log4j.properties
+++ b/bootstrap/src/integration-test/resources/log4j.properties
@@ -1,0 +1,4 @@
+log4j.rootLogger=INFO, X
+log4j.appender.X=org.apache.log4j.ConsoleAppender
+log4j.appender.X.layout=org.apache.log4j.PatternLayout
+log4j.appender.X.layout.ConversionPattern=[%d{dd.MM.yyyy HH:mm:ss,SSS}] bootstrap [%-5p] [%t] %m%n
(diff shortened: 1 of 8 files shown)


bootstrap/src/integration-test/resources/log4j.properties
```diff
@@ -0,0 +1,4 @@
+log4j.rootLogger=INFO, X
+log4j.appender.X=org.apache.log4j.ConsoleAppender
+log4j.appender.X.layout=org.apache.log4j.PatternLayout
+log4j.appender.X.layout.ConversionPattern=[%d{dd.MM.yyyy HH:mm:ss,SSS}] bootstrap [%-5p] [%t] %m%n
```
yurivin: May I ask you to use xml, json or yaml configuration instead of .properties?

Mingela: I think there should be created a separate refactoring task since we have the same properties file in all the modules

yurivin: All modules of notary, or for entire D3?

Mingela: Screen Shot 2019-04-05 at 10 58 59

PR #507, opened 2019-04-15
Improved Bitcoin wallet save mechanism

dolgopolovwork:
### Description of the Change
Bitcoin transactions must be saved in the wallet file in order to use them in the withdrawal service as UTXO. It's senseless to save every transaction in Bitcoin since it's too much. We used to save D3 clients deposit transactions only, but it's not enough. Change address transactions must be saved too.


configs/btc/deposit_local.properties
```diff
@@ -2,6 +2,8 @@
 btc-deposit.registrationAccount=btc_registration_service@notary
 btc-deposit.healthCheckPort=7070
 btc-deposit.btcTransferWalletPath=deploy/bitcoin/regtest/transfers.d3.wallet
+btc-deposit.mstRegistrationAccount=mst_btc_registration_service@notary
```
Mingela: why are there different `mst` and `non-mst` registration accounts?

Issue #508, opened 2019-04-16
Deploy Notary Smart contracts by separate Endpoints


yurivin:
Requested by Artyom.
While ETH network smart contract deploy is unstable because of various reasons such as fee changes, nonce unsynchrony e. t. c. it will help to to save funds if we will give opportunity to manually control sequence of smart contracts deployment for devops in Bootstrap service for case when something went wrong.

PR #509, opened 2019-04-16
- Github #508 Bootstrap ReimplementSmart contracts creation as separa…

yurivin:
- Github #508 Bootstrap ReimplementSmart contracts creation as separate methods to reduce ethereum payments.

### DIFF
--- a/bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/EthController.kt
+++ b/bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/EthController.kt
@@ -1,5 +1,6 @@
 package jp.co.soramitsu.bootstrap.controller
 
+import com.d3.eth.sidechain.util.DeployHelper
 import com.d3.eth.sidechain.util.DeployHelperBuilder
 import com.d3.eth.sidechain.util.hashToAddAndRemovePeer
 import jp.co.soramitsu.bootstrap.dto.*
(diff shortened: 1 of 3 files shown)


bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/EthController.kt
```diff
@@ -69,38 +70,91 @@ class EthController {
...
+            ResponseEntity.status(HttpStatus.CONFLICT)
+                .body(DeploySmartContractResponse(e.javaClass.simpleName, e.message))
+        }
+    }
+
+    @PostMapping("/deploy/D3/smartContracts")
+    fun deployInitialSmartContracts(@NotNull @RequestBody request: AllInitialContractsRequest): ResponseEntity<DeployInitialContractsResponse> {
         log.info { "Run predeploy with notary addresses: ${request.notaryEthereumAccounts}" }
-        var response: MasterContractResponse
+        var response: DeployInitialContractsResponse
```
Mingela: do we need this variable? I mean we can declare `val` before the `return`

Mingela (approved):

bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/EthController.kt
yurivin: yep

PR #514, opened 2019-04-22
MultiSig Iroha consumer

dolgopolovwork:
### Description of the Change
New MultiSigIrohaConsumer class that doesn't block caller thread until sent MSTs are fully committed.


notary-commons/src/main/kotlin/com/d3/commons/notary/NotaryImpl.kt
```diff
@@ -15,15 +16,17 @@ import java.math.BigInteger
 /**
  * Implementation of [Notary] business logic
  */
+
 class NotaryImpl(
-    private val notaryCredential: IrohaCredential,
```
Mingela: imho passing credential is more ideologically correct than retrieving key pair from a consumer, I do not like signatures like NotaryImpl(credential, consumer, events) or NotaryImpl(credential, irohaApi, ismultisig, events) neither because of redundancy but maybe they are slightly better

PR #515, opened 2019-04-22
Documentation for D3 eth notary initial smart contracts deployment sequence.

yurivin:


bootstrap/ReadMe.md
```diff
@@ -31,3 +31,13 @@ This function should return string json representatin of genesis block. It takes
...
+
+## Ethereum endpoint
+
+### Deploy smart contracts
+There are function to deploy all smart contracts by one API call: 'eth/deploy/D3/smartContracts' and functions to deploy initial
+ethereum contracts separately.
+Sequence to call smart contract deployment functions:
+1 - 'eth/deploy/D3/relayRegistry' - then take resulting contract address as an argument for the next function
+2 - 'eth/deploy/D3/masterContract' - then take resulting smart contract address as an argument to the next function
+3 - 'eth//deploy/D3/relayImplementation'
```
dolgopolovwork: I see extra slash in URL

bootstrap/ReadMe.md
```diff
@@ -31,3 +31,13 @@ This function should return string json representatin of genesis block. It takes
 * `fun getAccountsNeeded(): List<AccountPrototype>` - This function should return Accounts which have to present in genesis block to generate credentials for.
 
 Example realisation for D3: `genesis.d3.D3TestGenesisFactory`
+
+## Ethereum endpoint
+
+### Deploy smart contracts
+There are function to deploy all smart contracts by one API call: 'eth/deploy/D3/smartContracts' and functions to deploy initial
+ethereum contracts separately.
```
Mingela: there shouldn't be a line break

bootstrap/ReadMe.md
```diff
@@ -31,3 +31,13 @@ This function should return string json representatin of genesis block. It takes
...
+Sequence to call smart contract deployment functions:
```
Mingela:
```suggestion
There is the correct sequence of smart contract deployment functions calls:
```

yurivin: It is better to use 'Below' or 'Here' instead of 'There'. 
May you explain why my option does not fit?

bootstrap/ReadMe.md
yurivin: yep

bootstrap/ReadMe.md
yurivin: yep

bootstrap/ReadMe.md
Mingela: There can be many options to replace the phrase.
You can use `The sequence to call smart contract deployment functions is following:` or my previous suggestion etc. You missed `the` and a predicate (сказуемое).

yurivin: fixed

bootstrap/ReadMe.md
```diff
@@ -31,3 +31,12 @@ This function should return string json representatin of genesis block. It takes
 * `fun getAccountsNeeded(): List<AccountPrototype>` - This function should return Accounts which have to present in genesis block to generate credentials for.
 
 Example realisation for D3: `genesis.d3.D3TestGenesisFactory`
+
+## Ethereum endpoint
+
+### Deploy smart contracts
+There are function to deploy all smart contracts by one API call: 'eth/deploy/D3/smartContracts' and functions to deploy initial ethereum contracts separately.
```
Alexey-N-Chernyshov: I think it should be mentioned that only owner can call contracts later.

PR #520, opened 2019-04-24
Feature/Bootstrap remove mc address

yurivin:
The functionality of peer address removal from master contract


bootstrap/src/main/kotlin/jp/co/soramitsu/bootstrap/controller/EthController.kt
```diff
@@ -69,6 +97,12 @@ class EthController {
             )
         }
     }
+    private fun prepareTrxHash(removePeerAddress: String): String {
```
Mingela: Trx?

yurivin: means transaction

Mingela (approved):

PR #541, opened 2019-05-22
Raw keyPair data

dolgopolovwork:
### Description of the Change
According to the DevOps team improvements list, we must store key pair data directly in config files. No paths.

### DIFF
--- a/notary-commons/src/main/kotlin/com/d3/commons/config/Configs.kt
+++ b/notary-commons/src/main/kotlin/com/d3/commons/config/Configs.kt
@@ -44,6 +44,7 @@ interface IrohaConfig {
     val port: Int
 }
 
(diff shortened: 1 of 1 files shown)


notary-commons/src/main/kotlin/com/d3/commons/config/Configs.kt
```diff
@@ -44,6 +44,7 @@ interface IrohaConfig {
     val port: Int
 }
 
+//TODO this config must be removed
```
Alexey-N-Chernyshov: May be remove now?

dolgopolovwork: It's used.

PR #578, opened 2019-06-19
Added a way to register without brvs

Mingela:
Added config parameter to disable brvs grants in registration service


notary-registration/src/main/kotlin/com/d3/commons/registration/NotaryRegistrationStrategy.kt
```diff
@@ -119,53 +122,72 @@ class NotaryRegistrationStrategy(
...
+            IrohaCommand.CommandAddSignatory(
+                accountId,
+                publicKey
+            )
+        )
+        if (isBrvsEnabled) {
+            commands.add(
+                IrohaCommand.CommandSetAccountQuorum(
+                    accountId,
+                    2
```
dolgopolovwork: What is 2? Looks like magic constant.

notary-registration/src/main/kotlin/com/d3/commons/registration/NotaryRegistrationConfig.kt
```diff
@@ -26,8 +26,11 @@ interface NotaryRegistrationConfig {
     /** Iroha account to store clients */
     val clientStorageAccount: String
 
+    /** Property indicating if a brvs used in the system **/
+    val isBrvsEnabled: Boolean
+
     /** Iroha account of validator service */
-    val brvsAccount: String
+    val brvsAccount: String?
```
Alexey-N-Chernyshov: Do we really need it nullable?

Mingela: Otherwise there is an exception thrown if there is no 'brvsAccount' property specified which is quite honest if 'brvsEnabled' is set to `false`

notary-registration/src/main/kotlin/com/d3/commons/registration/NotaryRegistrationStrategy.kt
Mingela: Okay, moved it to const val

PR #588, opened 2019-06-24
Fixed domain priority

Mingela:
Before there was a way to pass an empty domain and registration wouldn't work when it was expected to


notary-commons/src/main/kotlin/com/d3/commons/registration/RegistrationServiceEndpoint.kt
```diff
@@ -103,6 +103,14 @@ class RegistrationServiceEndpoint(
         server.start(wait = false)
     }
 
+    private fun determineDomain(vararg candidates: String?) =
+        try {
+            candidates.first { domainCandidate -> !domainCandidate.isNullOrBlank() }!!
+        } catch (e: NoSuchElementException) {
+            logger.error("All passed domains were invalid: $candidates", e)
```
dolgopolovwork: I think it's a caller duty to log this exception.

PR #592, opened 2019-06-25
Only null check for pubkey

Mingela:


notary-commons/src/main/kotlin/com/d3/commons/registration/RegistrationServiceEndpoint.kt
```diff
@@ -133,9 +133,9 @@ class RegistrationServiceEndpoint(
         var reason = ""
         if (name.isNullOrEmpty()) reason = reason.plus("Parameter \"name\" is not specified. ")
         if (domain.isNullOrEmpty()) reason = reason.plus("Parameter \"domain\" is not specified. ")
-        if (pubkey.isNullOrEmpty()) reason = reason.plus("Parameter \"pubkey\" is not specified.")
+        if (pubkey == null) reason = reason.plus("Parameter \"pubkey\" is not specified.")
 
-        if (name.isNullOrEmpty() || domain.isNullOrEmpty() || pubkey.isNullOrEmpty()) {
+        if (name.isNullOrEmpty() || domain.isNullOrEmpty() || pubkey == null) {
```
dolgopolovwork: This predicate may changed to `reason.isNullOrEmpty()`. And I suggest you rename it to something more meaningful. Like `erroReason`.

PR #587, opened 2019-06-22
Feature/expansion remove logic

Alexey-N-Chernyshov:

PR #597, opened 2019-06-28
Release/0.4.3 snapshot

Alexey-N-Chernyshov:
### Description of the Change
rework hash in listeners (#596)
Feature/expansion remove logic (#587)
Fixed Genesis block (#595)
Added amin permissions (#594)
Updated genesis block (#593)
Only null check for pubkey (#592)
Making createTesterAccount() public (#591)


Mingela (approved):
You've got more commits in the PR description than needed :)

PR #647, opened 2019-08-27
Notary client provider + removed CLIENT_DOMAIN const

dolgopolovwork:
### Description of the Change
New provider called `NotaryClientProvider`. 
`CLIENT_DOMAIN` const is removed.


Alexey-N-Chernyshov (approved):

Mingela (changes requested):
- add '@' to tests accounts
- check k->v pair instead of just key
- transform accountId properly

Mingela (approved):

Mingela (approved):

PR #655, opened 2019-09-11
Feature/idempotent registration

Alexey-N-Chernyshov:
### Description of the Change
Check if account is already registered with given pubkey. In case it exists return OK response.


deploy/iroha/genesis.block
```diff
@@ -69,7 +69,8 @@
                     "can_set_quorum",
                     "can_grant_can_set_my_quorum",
                     "can_grant_can_add_my_signatory",
-                    "can_get_peers"
+                    "can_get_peers",
+                    "can_get_all_signatories"
```
dolgopolovwork, line 73: Don't forget to write a changelog for that

notary-registration/src/main/kotlin/com/d3/commons/registration/NotaryRegistrationStrategy.kt
```diff
@@ -46,16 +49,67 @@ class NotaryRegistrationStrategy(
...
+     */
+    fun isRegistered(
+        accountName: String,
+        domainId: String,
+        publicKey: String
+    ): Result<Boolean, Exception> = Result.of {
+        queryHelper.getSignatories("$accountName@$domainId")
+            .fold(
+                { signatories ->
+                    if (signatories.map { it.toLowerCase() }.contains(publicKey.toLowerCase()))
```
dolgopolovwork: I'd rather add braces.

dolgopolovwork (approved):

deploy/iroha/genesis.block
Alexey-N-Chernyshov: ok

PR #677, opened 2019-11-07
New data model in notifications

dolgopolovwork:
### Description of the Change
We need 2 more fields:
1) Event id
2) Event time

### DIFF
--- a/notifications/src/main/kotlin/com/d3/notifications/event/SoraEvents.kt
+++ b/notifications/src/main/kotlin/com/d3/notifications/event/SoraEvents.kt
@@ -12,62 +12,80 @@ import java.math.BigDecimal
  */
 
 /**
- * Sora event marker-interface
+ * Sora event
+ * @param id - event id
+ * @param time - event time
  */
-interface SoraEvent
+open class SoraEvent(val id: String, val time: Long)
 
-data class SoraDepositEvent(
+class SoraDepositEvent(
     val accountIdToNotify: String,
     val amount: BigDecimal,
     val assetName: String,
+    id: String,
+    time: Long,
     val from: String?
-) : SoraEvent {
+) : SoraEvent(id, time) {
     companion object {
-        fun map(transferNotifyEvent: TransferNotifyEvent): SoraDepositEvent {
-            return SoraDepositEvent(
-                transferNotifyEvent.accountIdToNotify,
-                transferNotifyEvent.amount,
-                transferNotifyEvent.assetName,
-                transferNotifyEvent.from
-            )
-        }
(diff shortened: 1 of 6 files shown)


notifications/src/main/kotlin/com/d3/notifications/event/SoraEvents.kt
```diff
@@ -12,62 +12,80 @@ import java.math.BigDecimal
  */
 
 /**
- * Sora event marker-interface
+ * Sora event
+ * @param id - event id
+ * @param time - event time
  */
-interface SoraEvent
+open class SoraEvent(val id: String, val time: Long)
```
Mingela, line 19: what is the difference between this and BasicEvent?

notifications/src/main/kotlin/com/d3/notifications/init/NotificationInitialization.kt
```diff
@@ -69,20 +67,22 @@ class NotificationInitialization(
...
-                else if (isEthRegistration(commandWithCreator)) {
-                    handleEthRegistrationEventNotification(command)
-                }
-                // Notify Bitcoin registration
-                else if (isBtcRegistration(commandWithCreator)) {
-                    handleBtcRegistrationEventNotification(command)
+        getSetAccountDetailTransactions(block)
+            .forEach { tx ->
+                tx.payload.reducedPayload.commandsList.forEach { command ->
+                    val commandWithCreator = CommandWithCreator(command, tx.payload.reducedPayload.creatorAccountId)
```
Mingela, line 73: move it to the `else`? since it is not used in the first `if`

notifications/src/main/kotlin/com/d3/notifications/init/NotificationInitialization.kt
```diff
@@ -289,25 +313,34 @@ class NotificationInitialization(
     // Handles transfer event notification
     private fun handleClientToClientEventNotification(
         transferAsset: Commands.TransferAsset,
-        fee: OperationFee?
+        tx: TransactionOuterClass.Transaction,
+        fee: TransferFee?
     ) {
+        val description: String = if (transferAsset.description == null) {
```
Mingela: `= transferAsset.description?: ""`?

Mingela (approved):

notifications/src/main/kotlin/com/d3/notifications/event/SoraEvents.kt
dolgopolovwork: We use `SoraEvent` to pass it to the Sora notification service. Think of it as the main DTO superclass to work with Sora. I know that `BasicEvent` is very similar, but I expect these classes to change in the future.

Mingela (approved):

PR #679, opened 2019-11-08
Handle failed registrations

dolgopolovwork:
### Description of the Change
Now we can handle failed registrations


notifications/src/main/kotlin/com/d3/notifications/init/NotificationInitialization.kt
```diff
@@ -77,6 +77,7 @@ class NotificationInitialization(
         getSetAccountDetailTransactions(block)
             .forEach { tx ->
                 tx.payload.reducedPayload.commandsList.forEach { command ->
+                    //TODO omg refactor it
```
Mingela, line 80: lyuto plyusuyu

Mingela (approved):

Alexey-N-Chernyshov (approved):

PR #694, opened 2019-11-18
Separate RMQ message for every NotificationService

dolgopolovwork:
### Description of the Change
We have to create an RMQ message for every notification service to handle them(messages) separately.


Mingela (approved):
How about using some enum instead of class names? it can be broken in case of inaccurate refactoring

Alexey-N-Chernyshov (approved):

PR #702, opened 2019-11-22
Add txIndex and blockNum to notifications

dolgopolovwork:
### Description of the Change
Sora needs to know transaction indexes, block numbers and tx time.

### DIFF
--- a/notifications/src/main/kotlin/com/d3/notifications/handler/CommandHandler.kt
+++ b/notifications/src/main/kotlin/com/d3/notifications/handler/CommandHandler.kt
@@ -5,9 +5,13 @@
 
 package com.d3.notifications.handler
 
+import iroha.protocol.BlockOuterClass
 import iroha.protocol.Commands
 import iroha.protocol.TransactionOuterClass
+import jp.co.soramitsu.iroha.java.Utils
 import mu.KLogging
+import java.util.*
+import java.util.stream.IntStream
 
 /**
  * The main command handler class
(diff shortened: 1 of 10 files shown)


Alexey-N-Chernyshov (approved):

notifications/src/main/kotlin/com/d3/notifications/handler/CommandHandler.kt
```diff
@@ -16,35 +19,60 @@ abstract class CommandHandler {
...
+    val tx: TransactionOuterClass.Transaction,
+    val block: BlockOuterClass.Block
+) {
+    /**
+     * Returns index of the current tx in the current block
+     */
+    fun getTxIndex(): Int {
+        var txIndex = 0
+        block.blockV1.payload.transactionsList.forEach {
+            if (Arrays.equals(Utils.hash(tx), Utils.hash(it))) {
```
Mingela: you don't need to compute tx hash each iteration.
is it legal to use `forEach` instead of classic iteration (maybe in Kotlin yes).
I'd propose
```
// Java
IntStream.range(0, transactionsList.size())
     .filter(i -> Arrays.equals(Utils.hash(transactionsList.get(i)), neededHash)
     .findFirst();
```
that would be an `optional` of index

Mingela (changes requested):
see the comment pls

Mingela (approved):

notifications/src/main/kotlin/com/d3/notifications/handler/CommandHandler.kt
dolgopolovwork: Done

